PR #46, opened 2018-06-13
Fix inconsistent test results in connector

AdityaKhatri:


src/views/Connector/index.js
```diff
@@ -260,6 +260,7 @@ export default class Connector extends React.PureComponent {
             <ConnectorDetails
                 className={styles.connectorDetails}
                 connectorId={connectorId}
+                key={connectorId}
```
tnagorra: Why is this key added?

PR #80, opened 2018-07-08
Make small changes to Export and Discover projects

bibekdahal:
* Add remaining columns to discover projects
* Make columns sortable in discover projects
* Add uncategorized section in report structure export
* Fix bug in `entryUpdater`

Requirements in server:
https://github.com/the-deep/server/pull/48
https://github.com/the-deep/server/pull/49


src/views/DiscoverProjects/index.js
```diff
@@ -196,7 +198,16 @@ export default class DiscoverProjects extends React.PureComponent {
                         mode="dd-MM-yyyy"
                     />
                 );
-            case 'analysis_framework_title':
+            case 'created_by':
+                return (
+                    <Link
+                        key={project.createdBy}
```
tnagorra: The key prop is not necessary here.

src/views/Export/ExportTypePane/index.js
```diff
@@ -74,6 +74,13 @@ export default class ExportTypePane extends React.PureComponent {
             });
         });
 
+        nodes.push({
+            title: 'Uncategorized',
```
tnagorra: Should we not translate the title?

PR #83, opened 2018-07-09
Add tooltip for assisted tagging

pprabesh:


src/components/AssistedTagging/styles.scss
```diff
@@ -34,6 +34,20 @@ $height-assistant-options: 58px;
             flex-basis: 40%;
             margin-left: $spacing-small;
         }
+
+        .info {
+            position: absolute;
+            top: 15px;
+            right: 60px;
```
AdityaKhatri: Always declare variables for sizes and use them.

tnagorra: Seconded

AdityaKhatri: Also change the right to 12px

PR #88, opened 2018-07-11
Feature select widget

tnagorra:


src/views/EditEntries/widgets/DateRange/index.js
```diff
@@ -0,0 +1,40 @@
...
+import styles from './styles.scss';
+
+const propTypes = {
+};
+
+const defaultProps = {
+};
+
+// eslint-disable-next-line react/prefer-stateless-function
+export default class DateRangeWidget extends React.PureComponent {
```
AdityaKhatri, line 14: Wouldn't a functional component be better here?

src/widgets/SelectWidget/Framework/ListWidget.js
```diff
@@ -0,0 +1,255 @@
...
+const defaultProps = {
+    data: {},
+};
+
+const emptyObject = {};
+const emptyList = [];
+
+@BoundError(WidgetError)
+export default class Select extends React.PureComponent {
+    static valueKeyExtractor = d => d.key;
```
AdityaKhatri: After default props

PR #90, opened 2018-07-12
Feature entries listing page

frozenhelium:


src/views/Entries/widgets/index.js
```diff
@@ -0,0 +1,75 @@
...
+            />
+        );
+    },
+);
+
+export const fetchWidget = widgetId => (
+    widgets[widgetId]
+);
+
+export const dummy = {};
```
AdityaKhatri: why dummy export?

src/views/Entries/Entry/index.js
```diff
@@ -0,0 +1,108 @@
...
+    className: '',
+};
+
+const widgetKeySelector = widget => widget.key;
+
+export default class Entry extends React.PureComponent {
+    static propTypes = propTypes;
+    static defaultProps = defaultProps;
+
+    layoutSelector = (widget) => {
```
AdityaKhatri: Why not static?
No this used, doesn't belong here

PR #95, opened 2018-07-13
Add excerpt widget for entries listing page

frozenhelium:


src/views/Entries/widgets/Excerpt/index.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+const defaultProps = {
+    className: '',
+    excerpt: '',
+    image: '',
+};
+
+const TEXT = 'excerpt';
+// const IMAGE = 'image';
```
tnagorra: Entry type must be either TEXT or IMAGE

src/widgetComponents/ExcerptOutput/index.js
```diff
@@ -0,0 +1,84 @@
...
+            ${classNameFromProps}
+            ${styles.excerpt}
+        `;
+
+        const Text = this.renderText;
+        const Image = this.renderImage;
+
+        return (
+            <div className={className}>
+                <Text />
```
tnagorra: use if else here instead of checking null inside individual render

PR #99, opened 2018-07-15
Add Matrix2dListOutput component

tnagorra:


src/views/Entries/widgets/Matrix2dList/index.js
```diff
@@ -0,0 +1,57 @@
...
+        const {
+            className,
+            data,
+            widget,
+        } = this.props;
+
+        const dimensions = getDimensions(widget);
+        const sectors = getSectors(widget);
+
+        const { value } = data;
```
AdityaKhatri, line 46: Deconstruct during access from this.props

PR #113, opened 2018-07-18
Add additional error message for recaptcha

thenav56:
Merge Required: https://github.com/the-deep/server/pull/58


tnagorra (commented):
Same thing needs to be done for Register page.
Also, captch is not shown

image

PR #115, opened 2018-07-19
Feature excerpt box geo organigram

AdityaKhatri:


src/components/OrganigramInput/index.js
```diff
@@ -206,6 +209,11 @@ export default class OrganigramInput extends React.PureComponent {
                     {/*
                         All organigrams in widgets have singular head
                     */}
+                    {modalLeftComponent &&
```
tnagorra, line 212: This code block should be before the comment above

PR #133, opened 2018-07-29
Feature refactor requests frameworks

AdityaKhatri:


src/redux/reducers/siloDomainData/analysisFramework.js
```diff
@@ -149,17 +149,24 @@ const getValidatedAnalysisFramework = (analysisFramework) => {
     // Return updates widget list
     return update(analysisFramework, {
         widgets: { $set: widgets },
+        pristine: { $set: true },
```
tnagorra: The second argument to update should be written as a separate variable.

src/views/AnalysisFramework/index.js
```diff
@@ -67,6 +61,20 @@ export default class AnalysisFramework extends React.PureComponent {
...
+            setState: params => this.setState(params),
+            setAnalysisFramework: this.props.setAnalysisFramework,
+            getAnalysisFramework: () => this.props.analysisFramework,
+        });
+
+        this.frameworkSaveRequest = new FrameworkSaveRequest({
+            setState: params => this.setState(params),
+            setAnalysisFramework: this.props.setAnalysisFramework,
+        });
+        console.warn(this.props.analysisFramework);
```
tnagorra: Unnecessary console.warn detected.

src/views/AnalysisFramework/index.js
```diff
@@ -209,7 +153,9 @@ export default class AnalysisFramework extends React.PureComponent {
             projectId,
         } = this.props;
 
-        if (!analysisFramework) {
+        const { pendingFramework } = this.state;
+
+        if (!analysisFramework || pendingFramework) {
```
tnagorra: analysisFramework is set to {} in selector so !analysisFramework is always false.
Anyway, analysisFramework should not be used to identify pending state.

PR #152, opened 2018-08-05
Implement computer and first time value calculator in edit entries

bibekdahal:
Requirement in react-store: https://github.com/toggle-corp/react-store/pull/308

* When an entry is first created, information date is set to lead's published date in date widget
* In the computer schema, currently we only have the default value of scale widget but we can do more in future

### DIFF
--- a/src/views/EditEntries/entryDataCalculator.js
+++ b/src/views/EditEntries/entryDataCalculator.js
@@ -1,3 +1,5 @@
+import update from '#rsu/immutable-update';
+
 const emptyObject = {};
 
 const calculateMatrix1dColor = (value, widgetData) => {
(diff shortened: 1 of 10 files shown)


src/views/EditEntries/entryDataCalculator.js
```diff
@@ -60,7 +60,35 @@ const calculateEntryData = (values = {}, analysisFramework) => {
...
+
+        if (widgetId === 'dateWidget' && widgetData.informationDateSelected) {
+            value = lead.publishedOn;
+        }
+
+        if (value) {
+            acc[widget.id] = { data: { value } };
+        }
+        return acc;
+    }, attributes,
```
tnagorra: we should not mutate attributes

tnagorra (approved):

PR #153, opened 2018-08-06
Change data strucutre for widget metadata

tnagorra:
- Add widget list visibility according to current page
- Add widget visibility according to current page and where widget was
added
- Add visibility modes: show, hidden, readonly


src/views/AnalysisFramework/List/index.js
```diff
@@ -14,14 +15,14 @@ const propTypes = {
     analysisFramework: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
 };
 
-export default class List extends React.PureComponent {
-    static propTypes = propTypes;
+console.warn(widgetList);
```
frozenhelium: Remove unnecessary console.warn

PR #154, opened 2018-08-06
Feature improve entries listing

frozenhelium:


src/views/Entries/requests/FrameworkRequest.js
```diff
@@ -16,6 +16,7 @@ export default class FrameworkRequest extends Request {
     }
 
     handleSuccess = (response) => {
+        console.warn(response);
```
tnagorra: console.warn alert

PR #161, opened 2018-08-12
Fix null undefined problem

AdityaKhatri:


src/redux/reducers/domainData/regions.js
```diff
@@ -123,11 +123,10 @@ const removeProjectRegion = (state, action) => {
 
 const addNewRegion = (state, action) => {
     const { regionDetail, projectId } = action;
+    console.warn(regionDetail, projectId);
```
thenav56: Remove `console.warn`

src/rest/projects.js
```diff
@@ -53,6 +54,12 @@ export const createParamsForProjectPatch = data => ({
     body: JSON.stringify(data),
 });
 
+export const createParamsForProjectPut = data => ({
+    method: PATCH,
```
thenav56: Please inspect.
in createParamsForProjectPut, method is PATCH

thenav56 (approved):

PR #182, opened 2018-08-26
Add Confirm on AF Save

thenav56:
Merge Dep: https://github.com/toggle-corp/react-store/pull/335


src/widgets/index.js
```diff
@@ -373,25 +374,25 @@ export const widgetListingVisibility = (widgetId, ...otherParams) => {
 
 // Determine visibility of widgets for WidgetEditor
 export const widgetVisibility = (widgetId, ...otherParams) => {
-    const overviewWidgetFn = () => VISIBILITY.show;
     const listWidgetFn = (view, addedFrom) => {
```
bibekdahal: This is no longer `listWidgetFn`. Also since all widgets have same logic, following mapping is probably not needed.

PR #203, opened 2018-09-19
Fix entries save

tnagorra:


thenav56:
This has been already implemented.

PR #234, opened 2018-10-28
Fix parameter issue with onChange for matrix widgets

tnagorra:


src/widgetComponents/Matrix1dInput/index.js
```diff
@@ -46,7 +46,6 @@ export default class Matrix1dInput extends React.PureComponent {
     handleCellDrop = (key, cellKey, droppedData) => {
         const { type, data } = droppedData;
         this.props.onChange(
```
bibekdahal: This is unnecessary change as the problem was not here but with Faram.

In certain parts of our code, the second parameter of onChange is `faramError` and the third parameter is `faramInfo`. But in other parts, the second parameter is `faramInfo`.

I have updated the Faram with the fix in https://github.com/toggle-corp/react-store/compare/feature-tabular.

PR #237, opened 2018-10-30
Add permission role for adding/editing leads

tnagorra:
- Fix circular dependency for selectors/usergroups

### DIFF
--- a/src/redux/selectors/siloDomainData/addLeads.js
+++ b/src/redux/selectors/siloDomainData/addLeads.js
@@ -17,6 +17,10 @@ import {
     projectIdFromRoute,
     connectorIdFromRoute,
 } from '../domainData';
+import {
+    activeProjectRoleSelector,
+} from './common';
+
 
 const emptyList = [];
 const emptyObject = {};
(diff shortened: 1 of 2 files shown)


src/redux/selectors/siloDomainData/addLeads.js
```diff
@@ -130,16 +134,21 @@ export const addLeadViewLeadKeysSelector = createSelector(
...
 export const addLeadViewLeadStatesSelector = createSelector(
     // TODO: distribute this later
+    activeProjectRoleSelector,
     addLeadViewLeadRestsSelector,
     addLeadViewLeadUploadsSelector,
     addLeadViewLeadDriveRestsSelector,
     addLeadViewLeadDropboxRestsSelector,
     addLeadViewLeadsSelector,
-    (leadRests, leadUploads, leadDriveRests, leadDropboxRests, leads) => (
+    (projectRole, leadRests, leadUploads, leadDriveRests, leadDropboxRests, leads) => (
```
frozenhelium, line 143: Maybe consider it changing to one object instead of millions of arguments,
Use multiple lines

tnagorra: it's legacy code now. breaking this to multiple lines will only make it more confusing

PR #239, opened 2018-10-31
Fix key issue, organigram child issue in conditional widget

AdityaKhatri:


src/widgets/edit/Conditional/index.js
```diff
@@ -126,12 +126,10 @@ export default class ConditionalWidgetEdit extends React.PureComponent {
             widgetKey: key,
             createNewElement: widgetData => ({
                 widget: {
-                    key: `${widgetData.widgetId}-${randomString(16).toLowerCase()}`,
+                    key: `${widgetData.widgetId}-${randomString(16).toLowerCase()}`.toLowerCase(),
```
frozenhelium: Redundant intermediate toLowerCase()

PR #245, opened 2018-11-04
Add search filed in schema of workshop faram

AdityaKhatri:


src/views/Workshop/index.js
```diff
@@ -52,6 +52,7 @@ export default class Workshop extends React.PureComponent {
 
     static schema = {
         fields: {
+            search: [],
```
tnagorra, line 52: Fix? It was intentionally left out from schema :D

PR #244, opened 2018-11-04
Feature better permission struct

tnagorra:


src/redux/reducers/domainData/projects.js
```diff
@@ -204,18 +209,31 @@ const setUserProjects = (state, action) => {
...
+                    },
+                ),
             }
         ),
         { },
     );
     const settings = {
         projectRoles: { $set: projectRoles },
     };
+    console.warn(settings);
```
AdityaKhatri: what is this?

PR #255, opened 2018-11-20
Feature project activity log

AdityaKhatri:


src/views/Project/Details/General/Dashboard/index.js
```diff
@@ -33,7 +33,12 @@ const propTypes = {
 
 const defaultProps = {
     className: '',
-    projectDashboard: {},
+    projectDashboard: {
```
tnagorra: leadActivity and entriesActivity should be checked for undefined values where it is used

src/utils/common.js
```diff
@@ -20,3 +20,8 @@ export const pick = (obj, keys) => keys.reduce(
     (acc, key) => ({ ...acc, [key]: obj[key] }),
     {},
 );
+
+export const camelToNormalCase = (text) => {
```
tnagorra, line 26: Fails for case like "IAmBADBoy"

tnagorra: Also missing test for function in utils

src/components/ActivityLog/index.js
```diff
@@ -0,0 +1,57 @@
...
+    render() {
+        const {
+            log,
+            className,
+        } = this.props;
+
+        return (
+            <div className={`${styles.activityLog} ${className}`}>
+                <h3 className={styles.header}>
+                    Activity Log
```
tnagorra: Every string should be translated

src/components/ActivityLog/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default class ActivityLog extends React.PureComponent {
+    static propTypes = propTypes;
+    static defaultProps = defaultProps;
+
+    static logItemKeySelector = l => l.key;
+
+    constructor(props) {
+        super(props);
+        this.state = {};
```
tnagorra: Unused this.state

src/components/ActivityLog/LogItem/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+import AggregatedLogItem from '../AggregatedLogItem';
+import styles from './styles.scss';
+
+const LogItem = ({
+    user,
+    time,
+    fields,
+}) => {
+    if (Object.keys(fields).length > 1) {
```
tnagorra, line 20: A comment would help

src/utils/common.js
AdityaKhatri: no!

PR #346, opened 2018-12-02
Use listSelection with segment prop in discover projects

AdityaKhatri:
https://github.com/toggle-corp/react-store/pull/397


src/views/DiscoverProjects/FilterProjectsForm.js
```diff
@@ -180,14 +180,15 @@ export default class FilterProjectsForm extends React.PureComponent {
                     showLabel
                     className="projects-filter"
                 />
-                <MultiSelectInput
+                <ListSelection
```
tnagorra: MultiSegmentInput use this

PR #352, opened 2018-12-03
Add an overlay for non droppable widgets in edit entries page

frozenhelium:


src/views/EditEntries/List/WidgetFaramContainer/index.js
```diff
@@ -109,7 +112,9 @@ export default class WidgetFaramContainer extends React.PureComponent {
         `;
 
         return (
-            <div className={className}>
+            <div
```
tnagorra, line 115: Line farming

PR #396, opened 2018-12-11
Add genericOnFailure function

AdityaKhatri:


frozenhelium:
`genericOnFailure` ? There is no trace of that function. Maybe you meant `notifyOnFailure`

PR #589, opened 2019-01-08
Add lead preview component in Leads table

pprabesh:


src/views/Leads/index.js
```diff
@@ -396,6 +409,10 @@ export default class Leads extends React.PureComponent {
         );
     }
 
+    handleLeadPreviewClose = () => {
+        this.setState({ showLeadPreview: false });
```
tnagorra: this should just call this.unsetLeadPreview instead

PR #650, opened 2019-01-27
Fix errored input border color in assessment

frozenhelium:


tnagorra:
Inconsistent color for date label
grim_2019-01-27-161131

PR #699, opened 2019-02-25
Fix loading icon behavior inside modal in Tabular Extraction

pprabesh:


src/views/LeadAdd/LeadFormItem/LeadTabular/index.js
```diff
@@ -272,20 +274,22 @@ export default class LeadTabular extends React.PureComponent {
...
-            return 'pending';
-        }
         if (responseError) {
-            return 'error';
+            notify.send({
+                title: _ts('addLeads.tabular', 'tabularExtractionError'),
+                type: notify.type.SUCCESS,
+                message: _ts('addLeads.tabular', 'tabularExtractionError'),
+                duration: notify.duration.MEDIUM,
+            });
```
AdityaKhatri: Will throw an error as this is a non blocking task.
Must close the modal.

PR #836, opened 2019-07-23
Pull one admin level at a time

AdityaKhatri:


src/components/geo/RegionMap/index.js
```diff
@@ -192,9 +192,11 @@ export default class RegionMap extends React.PureComponent {
     handleAdminLevelSelection = (id) => {
         this.setState({
             selectedAdminLevelId: id,
-        });
+        }, () => this.loadGeoJsons());
```
tnagorra: Why was the function not called directly?

tnagorra (changes requested):
Needs some changes

PR #839, opened 2019-07-30
Feature/add private badge user usergroup page

AdityaKhatri:
Adds badge on private projects on project lists in user profile page and user group page
Discover projects' name z-index is updated to fix the bug regarding overflow of name above table


frozenhelium:
Please add the details & changes by this pull request

PR #850, opened 2019-08-07
Fetch geojson from S3

AdityaKhatri:
Depends on https://github.com/the-deep/server/pull/358

### DIFF
--- a/src/components/geo/GeoViz/Region.js
+++ b/src/components/geo/GeoViz/Region.js
@@ -18,8 +18,6 @@ import {
     RequestCoordinator,
 } from '#request';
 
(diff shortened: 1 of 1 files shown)


src/components/geo/GeoViz/Region.js
```diff
@@ -48,9 +55,9 @@ const requests = {
     regionRequest: {
         onMount: true,
         onPropsChanged: ['regionId'],
-        url: ({ props: { regionId } }) => `/regions/${regionId}/`,
+        url: ({ props: { regionId } }) => createUrlForAdminLevelsForRegion(regionId),
```
tnagorra: Let us not use this pattern for url creation

tnagorra (approved):

PR #857, opened 2019-08-16
Fix/date filters

AdityaKhatri:
Depends on: https://github.com/the-deep/server/pull/373/files
Fixes published data, created date filter in leads page
Fixes created date filter in entries listing page

### DIFF
--- a/src/utils/common.js
+++ b/src/utils/common.js
@@ -1,3 +1,5 @@
+import { padStart } from '@togglecorp/fujs';
+
 export const mapObjectToObject = (obj, fn) => {
     const newObj = {};
     Object.keys(obj).forEach((key) => {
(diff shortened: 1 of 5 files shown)


src/utils/common.js
```diff
@@ -79,3 +79,19 @@ export const timeFrom = (date) => {
...
     return 'just now';
 };
+
+export const getDateWithTimezone = (date) => {
+    const newDate = new Date();
+    const timezoneOffset = newDate.getTimezoneOffset();
+    const reversesign = timezoneOffset < 0 ? '+' : '-';
+
+    const hours = Math.floor(Math.abs(timezoneOffset) / 60);
+    const hoursString = hours < 10 ? `0${hours}` : hours;
```
tnagorra: Use padString from fujs

src/utils/common.js
```diff
@@ -79,3 +79,19 @@ export const timeFrom = (date) => {
...
+
+    const minutes = Math.abs(timezoneOffset) % 60;
+    const minutesString = minutes < 10 ? `0${minutes}` : minutes;
```
tnagorra: Use padString from fujs

PR #873, opened 2019-09-15
 Feature/lead create

tnagorra:


src/views/LeadAdd/LeadPreview/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.tabular-modal {
+    transition: .2s all ease-in-out;
+    width: auto;
+    overflow: auto;
+
+    :global {
+        .page-1 {
+            width: 50vw;
```
frozenhelium, line 36: should use variables from `styleProperties`

PR #892, opened 2019-09-20
Disable form if user is not admin

AdityaKhatri:
* Add migration in redux to remove siloDomaindata.connectorsView

Merge: https://github.com/toggle-corp/react-store/pull/649


src/views/Connector/Details/Form/connector-utils.js
```diff
@@ -68,6 +68,7 @@ export const getUsersTableHeader = (handleToggleUserRoleClick, handleDeleteUserC
         order: 5,
         modifier: (row) => {
             const isAdmin = row.role === 'admin';
+            const isSelf = row.user === userId;
```
tnagorra, line 76: I don't think we should do this, as this action can be done through select input

src/components/other/FilterLeadsForm/index.js
```diff
@@ -56,12 +56,14 @@ const propTypes = {
 
     // eslint-disable-next-line react/forbid-prop-types
     leadOptionsRequest: PropTypes.object.isRequired,
+    onlyUnprotected: PropTypes.bool,
```
frozenhelium: A better prop name would be more comprehensive

src/views/Export/index.js
```diff
@@ -71,11 +79,68 @@ const propTypes = {
...
+        method: requestMethods.POST,
+        onMount: true,
+        query: ({
+            fields: ['id', 'title', 'created_at'],
+        }),
+        body: ({
+            props: {
+                projectId,
+                filters,
+                projectRole: {
```
frozenhelium, line 99: Is project role always defined?

src/views/Export/index.js
```diff
@@ -71,11 +79,68 @@ const propTypes = {
...
+                    exportPermissions = {},
+                },
+            },
+        }) => {
+            const onlyUnprotected = exportPermissions.create_only_unprotected;
+            const sanitizedFilters = getFiltersForRequest(filters);
+
+            if (onlyUnprotected) {
```
frozenhelium: We should add a comment here, no?

src/views/Export/index.js
AdityaKhatri: yes

src/views/Export/index.js
AdityaKhatri: No

src/views/Export/index.js
```diff
@@ -455,8 +459,15 @@ export default class Export extends React.PureComponent {
...
             geoOptions,
+            projectRole: {
+                exportPermissions = {},
+            },
+            leadsGetRequest: {
+                pending: pendingLeads,
+            },
         } = this.props;
         const { filters } = analysisFramework || {};
+        const onlyUnprotected = exportPermissions.create_only_unprotected;
```
frozenhelium: snake case looks ugly

AdityaKhatri: Clean your glasses

PR #903, opened 2019-10-04
Fix/sort roles in input

AdityaKhatri:

PR #904, opened 2019-10-21
Fix/list view stakeholder

AdityaKhatri:
First merge #903

Depends on https://github.com/toggle-corp/react-store/pull/667

### DIFF
--- a/src/views/EditAry/RightPanel/Metadata/Column/index.js
+++ b/src/views/EditAry/RightPanel/Metadata/Column/index.js
@@ -49,8 +49,8 @@ export default class Column extends React.PureComponent {
 
         return {
             ...widgetProps,
-            fieldType,
-            readOnly: isStakeholder,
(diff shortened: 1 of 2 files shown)


src/views/EditAry/RightPanel/Metadata/Column/index.js
```diff
@@ -49,8 +49,9 @@ export default class Column extends React.PureComponent {
 
         return {
             ...widgetProps,
-            fieldType,
-            readOnly: isStakeholder,
+            fieldType: isStakeholder ? 'listInput' : fieldType,
+            disabled: isStakeholder,
+            isStakeholder,
```
tnagorra: Why pass `isStakeholder` field?

tnagorra (changes requested):
1. The title looks disabled. I think it should still be readOnly.
2. We can still remove the elements even when disabled is true.
3. We shouldn't directly override the type of the widget.

PR #929, opened 2019-12-13
Add roated mode in matrix 1d

frozenhelium:
Note: not very preview friendly (i.e. may see different view when resizing the widget vs when actually rendered)


src/widgetComponents/Matrix1dInput/Row/index.js
```diff
@@ -30,20 +30,31 @@ const defaultProps = {
...
+
 export default class Matrix1dRow extends React.PureComponent {
     static propTypes = propTypes;
     static defaultProps = defaultProps;
 
     static cellKeyExtractor = d => d.key;
 
     rendererParams = (key, data) => ({
-        children: data.value,
+        children: (
```
tnagorra: We should avoid doing this. This will create an object in every render (removing the advantage of using List component)

src/widgets/edit/Matrix1d/index.js
```diff
@@ -51,18 +56,37 @@ const emptyMetaObject = {
...
+        image: matrix1dRotatedIcon,
     },
 ];
 
+const orientationOptionRendererParams = (_, d) => ({
+    image: d.image,
+    containerClassName: styles.orientationOption,
+});
+
+const OrientationOption = p => (
```
tnagorra, line 82: We should define propTypes for Components

tnagorra (changes requested):
Needs to fix some issues!

tnagorra (approved):

PR #936, opened 2019-12-19
Show ary documents in modal

AdityaKhatri:


src/components/input/MultiDocumentUploader/index.js
```diff
@@ -284,6 +285,7 @@ export default class MultiDocumentUploader extends React.PureComponent {
             && (isValueEmpty || isBeingDraggedOver)
             && !pending;
 
+        console.warn('here', value);
```
tnagorra: Let's not leave console.warn

PR #952, opened 2020-01-20
Cloak geo input polygon enable

AdityaKhatri:


src/components/input/GeoInput/index.js
```diff
@@ -202,6 +216,10 @@ export default class GeoInput extends React.PureComponent {
         const polygons = this.getPolygons(value);
         const options = this.getAllGeoOptions(geoOptionsByRegion);
 
+        const polygonSupportIndex = accessibleFeatures.findIndex(f => f.key === 'polygon_support_geo');
```
tnagorra: We should create constants for feature name somewhere

src/components/input/GeoInput/index.js
```diff
@@ -202,6 +216,10 @@ export default class GeoInput extends React.PureComponent {
...
+        const isPolygonFeatureEnabled = polygonSupportIndex !== -1;
+        const shouldEnablePloygon = isPolygonFeatureEnabled && polygonsEnabled;
```
tnagorra: Spelling mistake for "Polygon"
Try using this plugin for vim https://github.com/kamykn/spelunker.vim

tnagorra (changes requested):
Need to create accessible feature constants

src/components/input/GeoInput/index.js
AdityaKhatri: yes we should

AdityaKhatri: In constants?

src/constants/features.js
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line import/prefer-default-export
+const featuresMapping = {
+    polygonSupportGeo: 'polygon_support_geo',
+    zoomableImage: 'zoomable_image',
+    privateProject: 'privateProject',
```
tnagorra: The feature names from server are usually snake_case. This looks problematic

AdityaKhatri: It was a test. You passed!

tnagorra (approved):

PR #972, opened 2020-02-06
Add XForm export basic logic

thenav56:


src/typings/index.tsx
```diff
@@ -5,6 +5,10 @@ import {
 
 // Basic
 
+export interface StringStringElement {
```
tnagorra: Please use Obj<string> from @togglecorp/fujs

thenav56: Removed. Done

tnagorra (approved):

PR #976, opened 2020-02-07
Fix/builder styling

AdityaKhatri:


src/typings/questionnaire.tsx
```diff
@@ -81,6 +81,13 @@ export interface BaseQuestionElement {
     crisisType?: number;
     crisisTypeDetail?: IdTitle;
 
+    showCheckbox?: boolean;
```
tnagorra: This doesn't look right!

PR #1008, opened 2020-05-06
Add search filter to QB and framework questions

AdityaKhatri:


src/entities/questionnaire.ts
```diff
@@ -295,6 +301,20 @@ export function generateXLSForm(id: number, title: string, questions: BaseQuesti
     return workbook;
 }
 
+export function generateDurationLabel(duration?: number) {
+    if (isNotDefined(duration) || duration === 0) {
+        return '0 min';
```
tnagorra: We should show 'NA' if duration is not defined instead of '0 min'

src/entities/questionnaire.ts
```diff
@@ -295,6 +301,20 @@ export function generateXLSForm(id: number, title: string, questions: BaseQuesti
...
+    }
+    const durationSec = duration ? duration % 60 : 0;
```
tnagorra: Why check for duration here?

src/entities/questionnaire.ts
```diff
@@ -295,6 +301,20 @@ export function generateXLSForm(id: number, title: string, questions: BaseQuesti
...
+    const durationMin = duration
+        ? ((duration - durationSec) / 60) : 0;
+
+    const durationMinuteLabel = `${durationMin} min`;
+    const durationSecondLabel = durationSec === 0 ? '' : `${durationSec} sec`;
+
+    return `${durationMinuteLabel} ${durationSecondLabel}`;
```
tnagorra: Should be a bit smarter than this. Only show what is necessary.

src/entities/questionnaire.ts
```diff
@@ -162,16 +171,35 @@ export function getFrameworkMatrices(
...
+    if (!questions || (values.length <= 0 && isFalsyString(searchValue))) {
         return questions;
     }
 
-    const filteredQuestions = questions.filter(question => (
-        question.frameworkAttribute
-        && question.frameworkAttribute.value
-        && values.includes(question.frameworkAttribute.value)
-    ));
+    const filteredQuestions = questions.filter((question) => {
```
tnagorra, line 186: I think this can be simplified a bit.

src/entities/questionnaire.ts
```diff
@@ -295,6 +323,27 @@ export function generateXLSForm(id: number, title: string, questions: BaseQuesti
     return workbook;
 }
 
+export function generateDurationLabel(duration?: number) {
+    if (isNotDefined(duration) || duration === 0) {
```
tnagorra: Doesn't `duration === 0` mean the duration is `0 sec`?

src/components/questionnaire-builder/QuestionList/index.tsx
```diff
@@ -121,23 +129,65 @@ const QuestionList = (props: QuestionListProps) => {
...
+    }, [frameworkOptions, questions]);
+
     const filteredQuestions = useMemo(
         () => {
-            if (!questions) {
+            if (!flatQuestions) {
                 return [];
             }
-            return questions.filter(question => !!question.isArchived === archived);
+            return flatQuestions.filter((question) => {
```
tnagorra: Isn't this repeated here?

PR #1009, opened 2020-05-14
Feature/new remap use

tnagorra:
Depends on https://github.com/toggle-corp/re-map/pull/2
Depends on https://github.com/toggle-corp/react-store/pull/722

### DIFF
--- a/src/components/input/GeoInput/GeoModal/GeoInputList/index.js
+++ b/src/components/input/GeoInput/GeoModal/GeoInputList/index.js
@@ -9,7 +9,6 @@ import ListView from '#rscv/List/ListView';
 
 import styles from './styles.scss';
 
-// TODO: should handle multiple regions later
 const GeoInputList = (props) => {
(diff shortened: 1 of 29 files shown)


src/components/input/GeoInput/GeoModal/GeoInputList/index.js
```diff
@@ -1,24 +1,103 @@
...
+import { _cs, isNotDefined, isDefined } from '@togglecorp/fujs';
 
 import DismissableListItem from '#rsca/DismissableListItem';
 import ListView from '#rscv/List/ListView';
 
-
 import styles from './styles.scss';
 
-// TODO: should handle multiple regions later
+function groupList(
```
AdityaKhatri, line 42: Don't we already have this?

tnagorra: No. This is a bit different!

PR #1050, opened 2020-07-27
Add has assessment filter and indicator in leads page

AdityaKhatri:
Depends on:
https://github.com/the-deep/server/pull/478

### DIFF
--- a/src/views/LeadAdd/index.js
+++ b/src/views/LeadAdd/index.js
@@ -400,6 +400,7 @@ class LeadAdd extends React.PureComponent {
                     // NOTE: hard-coded priority value
                     priority: 'low',
                     // Inject other values too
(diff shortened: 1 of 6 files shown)


src/views/LeadAdd/index.js
```diff
@@ -397,7 +397,7 @@ class LeadAdd extends React.PureComponent {
                     publishedOn: formatDateToString(now, 'yyyy-MM-dd'),
                     // NOTE: hard-coded confidentiality value
                     confidentiality: 'unprotected',
-                    // Inject other values too
+                    authors: isDefined(faramValues.authors) ? faramValues.authors : [],
```
tnagorra, line 403: Why was authors added here? And why not other values like publishers?

AdityaKhatri: The field is optional, when passed null, it throws an error

PR #1053, opened 2020-07-30
Fix leads priority values

AdityaKhatri:


src/views/LeadAdd/index.js
```diff
@@ -398,7 +398,7 @@ class LeadAdd extends React.PureComponent {
                     // NOTE: hard-coded confidentiality value
                     confidentiality: 'unprotected',
                     // NOTE: hard-coded priority value
-                    priority: 'low',
+                    priority: 100,
```
tnagorra, line 401: It would be even better if we get this from leadOptions.priority[0]
Setting this here would be problematic.
Or, just not fill it altogether

AdityaKhatri: Fixed

src/views/LeadAdd/LeadDetail/index.js
```diff
@@ -402,6 +390,10 @@ class LeadDetail extends React.PureComponent {
         webInfoDataRequest.setDefaultParams({ handleWebInfoFill: this.handleWebInfoFill });
     }
 
+    getPriorityOptions = memoize((priority = []) => (
+        priority.sort((a, b) => compareNumber(a.key, b.key))
```
tnagorra: We should make a copy before sorting it

src/views/LeadAdd/LeadDetail/index.js
```diff
@@ -535,19 +527,29 @@ class LeadDetail extends React.PureComponent {
...
             }));
         }
 
-        /*
         const values = leadFaramValuesSelector(lead);
-        const newValues = fillExtraInfo(values, leadOptions, activeUserId);
-        this.handleLeadValueChange(newValues);
-        */
+        if (isNotDefined(values.priority) && isDefined(priority)) {
+            const sortedPriority = priority.sort((a, b) => compareNumber(a.key, b.key));
```
tnagorra: We should make a copy before sorting a list

src/views/LeadAdd/LeadDetail/index.js
```diff
@@ -535,19 +527,29 @@ class LeadDetail extends React.PureComponent {
...
+            const newValues = {
```
tnagorra: why was fillExtraInfo be removed?

PR #1058, opened 2020-08-04
Add fullscreen option while taking lead screenshot

AdityaKhatri:


src/components/viewer/GalleryViewer/index.js
```diff
@@ -155,9 +159,26 @@ export default class GalleryViewer extends React.PureComponent {
...
+        this.viewerRef = React.createRef();
+    }
+
+    componentDidMount() {
+        document.addEventListener('fullscreenchange', this.handleFullscreenChange);
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('fullscreenchange', this.handleFullscreenChange);
+    }
```
tnagorra: Why do this?
isFullscreen is set as false initially.
isFullscreen's value is irrelevant when a component dismounts.

AdityaKhatri: This is to remove the event listener from document

tnagorra (approved):

PR #1071, opened 2020-08-27
Refactor ExternalGallery to handle various error states

tnagorra:
- ErrorGallery now accepts 'error' props to show parent errors
- GalleryDocs uses embed instead of iframe for pdf

### DIFF
--- a/src/components/other/ExportPreview/index.js
+++ b/src/components/other/ExportPreview/index.js
@@ -125,10 +125,11 @@ export default class ExportPreview extends React.PureComponent {
                 <GalleryViewer
                     url={exportObj.file}
                     mimeType={exportObj.mimeType}
-                    // NOTE: should not this be true
+                    // NOTE: should not this be true?
                     canShowIframe={false}
-                    invalidUrlMessage={_ts('components.exportPreview', 'previewNotAvailableLabel')}
                     showUrl
+                    // eslint-disable-next-line max-len
+                    // invalidUrlMessage={_ts('components.exportPreview', 'previewNotAvailableLabel')}
                 />
             );
         }
--- a/src/components/viewer/GalleryViewer/GalleryDocs/index.js
+++ b/src/components/viewer/GalleryViewer/GalleryDocs/index.js
@@ -12,7 +12,10 @@ const propTypes = {
     className: PropTypes.string,
     docUrl: PropTypes.string,
     mimeType: PropTypes.string,
+
+    // can the docUrl be shown in iframe
     canShowIframe: PropTypes.bool,
+    // is the docUrl https
     notHttps: PropTypes.bool,
 };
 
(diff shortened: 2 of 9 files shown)


src/components/other/ExportPreview/index.js
```diff
@@ -125,10 +125,11 @@ export default class ExportPreview extends React.PureComponent {
                 <GalleryViewer
                     url={exportObj.file}
                     mimeType={exportObj.mimeType}
-                    // NOTE: should not this be true
+                    // NOTE: should not this be true?
                     canShowIframe={false}
```
tnagorra, line 129: @thenav56  `canShowIframe` for documents generated in deep should be `true`, right?

src/components/viewer/GalleryViewer/GalleryDocs/index.js
```diff
@@ -27,46 +30,51 @@ const defaultProps = {
...
+                    src={docUrl}
                 />
             </div>
         );
     }
+
+    return (
+        <div className={className}>
+            <iframe
+                className={`doc ${styles.doc}`}
```
AdityaKhatri: use _cs

src/components/viewer/GalleryViewer/GalleryDocs/index.js
```diff
@@ -27,46 +30,51 @@ const defaultProps = {
...
+                title={docUrl}
+                src={createUrlForGoogleViewer(docUrl)}
+                sandbox="allow-scripts allow-same-origin allow-popups"
```
AdityaKhatri, line 72: @tnagorra , aren't we removing these?

src/components/viewer/GalleryViewer/index.js
```diff
@@ -24,72 +24,84 @@ import GalleryDocs from './GalleryDocs';
...
     } else if (galleryMapping[mimeType] === galleryType.DOC) {
+        // NOTE: no need to check for https for GalleryDocs as it has google viewer as fallback
         return (
             <GalleryDocs
                 className={className}
                 docUrl={url}
                 mimeType={mimeType}
                 canShowIframe={canShowIframe}
                 notHttps={!isHttps}
-                {...otherProps}
```
AdityaKhatri, line 84: isn't this required?

src/components/viewer/GalleryViewer/index.js
```diff
@@ -129,26 +140,52 @@ const Bar = ({ url = '', children }) => {
                 showHintAndError={false}
                 selectOnFocus
             />
-            {
```
AdityaKhatri, line 132: why was this even here? hahaha

src/components/viewer/GalleryViewer/GalleryDocs/index.js
tnagorra: Didn't want to refactor old component :D

src/components/viewer/GalleryViewer/GalleryDocs/index.js
tnagorra: No idea

src/components/viewer/GalleryViewer/index.js
tnagorra: Nop, there is no remaining prop that galleryDocs accept.
I don't think we should let user override GalleryDocs props from GalleryViewer props anyway.

src/components/viewer/GalleryViewer/index.js
tnagorra: IDK

src/components/other/ExportPreview/index.js
thenav56: Yeah

PR #1079, opened 2020-09-03
Feature/lead state for candidate leads

AdityaKhatri:


src/views/LeadAdd/CandidateLeads/CandidateLeadsModal/index.js
```diff
@@ -1,6 +1,9 @@
 import React, { useCallback, useContext, useMemo } from 'react';
 import PropTypes from 'prop-types';
-import { _cs } from '@togglecorp/fujs';
+import {
+    _cs,
+    isDefined,
+} from '@togglecorp/fujs';
```
tnagorra: Is isDefined used?

src/views/LeadAdd/LeadProcessor/index.js
```diff
@@ -90,7 +91,22 @@ function LeadProcessor(props) {
...
+                const currentLeadIndex = safeCandidateLeads
+                    .findIndex(lead => leadKeySelector(lead) === key);
+                if (currentLeadIndex !== -1) {
+                    if (progress === 100) {
+                        // eslint-disable-next-line no-param-reassign
+                        safeCandidateLeads[currentLeadIndex].leadState = LEAD_STATUS.complete;
+                    } else {
+                        // eslint-disable-next-line no-param-reassign
+                        safeCandidateLeads[currentLeadIndex].leadState = LEAD_STATUS.uploading;
+                    }
```
tnagorra: IMO ternary operators are better here

src/views/LeadAdd/LeadProcessor/index.js
```diff
@@ -104,6 +120,18 @@ function LeadProcessor(props) {
...
+        setCandidateLeads(currentCandidateLeads => (
+            produce(currentCandidateLeads, (safeCandidateLeads) => {
+                const currentLeadIndex = safeCandidateLeads
+                    .findIndex(lead => leadKeySelector(lead) === key);
+                if (currentLeadIndex !== -1) {
+                    if (pending) {
+                        // eslint-disable-next-line no-param-reassign
+                        safeCandidateLeads[currentLeadIndex].leadState = LEAD_STATUS.uploading;
+                    }
+                }
```
tnagorra: why not just use `currentLeadIndex !== -1 && pending`

src/views/LeadAdd/LeadProcessor/index.js
```diff
@@ -178,6 +208,7 @@ function LeadProcessor(props) {
                     error: false,
                     pristine: isDefined(serverId),
                 },
+                leadState: LEAD_STATUS.pending,
```
tnagorra: Why set it to pending? The word pending is a bit misleading.
The state is pending. But, LEAD_STATUS.pending is actually working.
We should only set it to pending when the request is being made!

PR #1097, opened 2020-09-17
Add draw over screenshot feature

frozenhelium:
@tnagorra @AdityaKhatri good enough?


src/components/general/CanvasDraw/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+interface Props {
+    className?: string;
+    imgSrc?: string;
+    onDone?: (imgData: string) => void;
+}
+
+function CanvasDraw(props: Props) {
+    const {
+        className,
+        imgSrc = 'https://i.imgur.com/TjbE3JO.jpg',
```
AdityaKhatri: remove default value

src/components/general/CanvasDraw/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+                    />
+                    <SegmentInput
+                        // FIXME: use strings
+                        label="Pen size"
+                        options={penSizeOptions}
+                        value={penSize}
+                        onChange={handlePenSizeInputChange}
+                    />
+                    <Button onClick={handleClearButtonClick}>
+                        {/* FIXME: use strings */}
```
AdityaKhatri: Please use strings instead of adding a fixme

src/components/general/CanvasDrawModal/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+    const {
+        className,
+        imgSrc,
+        onDone,
+        onCancel,
+    } = props;
+    const canvasRef = React.useRef<HTMLCanvasElement>(null);
+    const canvasContainerRef = React.useRef<HTMLDivElement>(null);
+    const [color, setColor] = React.useState(
+        getComputedStyle(document.body).getPropertyValue('--color-accent'),
```
tnagorra: Use `useState(() => the value to be computed)` syntax

src/components/general/CanvasDrawModal/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+            } else {
+                canvas.width = bcr.width;
+                canvas.height = bcr.height;
+            }
+        }
+    }, [imgSrc]);
+
+    return (
+        <Modal className={styles.components.canvasDrawModal}>
+            <ModalHeader title="Draw over the screenshot" />
```
tnagorra: We should translate this as well

src/components/general/CanvasDrawModal/index.tsx
frozenhelium: but why?

src/components/general/CanvasDrawModal/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+                imageRef.current = img;
+            } else {
+                canvas.width = bcr.width;
+                canvas.height = bcr.height;
+            }
+        }
+    }, [imgSrc]);
+
+    return (
+        <Modal className={styles.components.canvasDrawModal}>
```
AdityaKhatri: What is `styles.components.canvasDrawModal`?

src/components/general/CanvasDrawModal/index.tsx
tnagorra: Because, the `getComputedStyle(document.body).getPropertyValue('--color-accent')` will run on every render.
We can avoid that using this syntax:
```suggestion
        () => getComputedStyle(document.body).getPropertyValue('--color-accent'),
```

tnagorra (approved):

PR #1098, opened 2020-09-17
Add export to image feature for framework

frozenhelium:


src/views/Project/Details/Frameworks/FrameworkDetail/index.js
```diff
@@ -127,6 +130,59 @@ const projectRendererParams = (_, p) => ({
...
+
+        document.body.appendChild(link);
+
+        link.click();
+
+        document.body.removeChild(link);
+    } else {
+        window.open(uri);
+    }
+}
```
tnagorra, line 153: We should move this to utils

src/views/Project/Details/Frameworks/FrameworkDetail/index.js
```diff
@@ -127,6 +130,59 @@ const projectRendererParams = (_, p) => ({
...
+
+const getFrameworkSize = (framework) => {
+    let width = 0;
+    let listHeight = 0;
+    let overviewHeight = 0;
+
+    framework.widgets.forEach((w) => {
```
tnagorra: Prefer a `Math.max(...framework.widgets.map(function to get the extremes));` apporach

src/views/Project/Details/Frameworks/FrameworkDetail/index.js
```diff
@@ -186,6 +245,27 @@ export default class FrameworkDetail extends React.PureComponent {
...
+        } = this.props;
+
+        this.setState({
+            isExporting: true,
+        }, () => {
+            setTimeout(() => {
+                if (this.exportSectionRef.current) {
+                    html2canvas(this.exportSectionRef.current).then((canvas) => {
+                        saveAs(canvas.toDataURL(), `export-${framework.title}.png`);
+                        this.setState({ isExporting: false });
```
tnagorra: This may be problematic as the async function will be called even after the component is unmounted.
We may need to look into how this can be safely done!

src/views/Project/Details/Frameworks/FrameworkDetail/index.js
frozenhelium: this is very custom tailored to this particular case

PR #1123, opened 2020-09-25
Add entry filter by lead properties

frozenhelium:
* Lead published on
* Lead confidentiality
* Lead priority

Depends on: https://github.com/the-deep/server/pull/504

### DIFF
--- a/src/entities/entries.js
+++ b/src/entities/entries.js
@@ -94,11 +94,11 @@ export const processEntryFilters = (filters, framework, geoOptions) => {
 
             result.push([
                 `${filterKey}__gte`,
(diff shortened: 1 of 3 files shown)


tnagorra (changes requested):
We need to disable these filters on Export page

frozenhelium:
@tnagorra Export page uses same component, so it should be automatically displayed there

AdityaKhatri:
@frozenhelium , what @tnagorra has mentioned is correct but as this is an urgent fix, let's create a separate issue for removing lead specific filters from the export page, and deploy this asap.

PR #1125, opened 2020-09-28
Add multiple lead selection in connector leads

AdityaKhatri:


src/hooks/stateManagement.ts
```diff
@@ -69,3 +73,45 @@ export function useModalState(initialValue: boolean): [
     return [visible, setVisible, setHidden, setVisibility];
 }
 
+export function useArraySelection<T, K extends string | number>(
+    keySelector: (value: T) => K,
+    defaultValue: T[] | undefined = [],
```
tnagorra: If we pass default value, then how can it be undefined?

src/hooks/stateManagement.ts
```diff
@@ -69,3 +73,45 @@ export function useModalState(initialValue: boolean): [
...
+            newValues.splice(index, 1);
+            return newValues;
+        });
+    };
+
+    const clearSelection = () => {
+        setValues([]);
+    };
+
+    const valuesMap = listToMap(values, keySelector, d => d);
```
tnagorra: This should be memoized!

frozenhelium (approved):

PR #1145, opened 2020-10-02
Add container for entry QC

frozenhelium:


src/views/Entries/index.js
```diff
@@ -271,19 +282,26 @@ export default class Entries extends React.PureComponent {
 
     getTabs = memoize((framework, isVisualizationEnabled) => {
         if (isVisualizationEnabled && isVisualizationEnabled.entry) {
+            const tabs = isDev ? {
```
AdityaKhatri: Write this for alpha as well

PR #1147, opened 2020-10-04
Add basic entry card in the qc page

frozenhelium:
Depends on: https://github.com/the-deep/server/pull/517

### DIFF
--- a/src/components/viewer/DateOutput/index.tsx
+++ b/src/components/viewer/DateOutput/index.tsx
@@ -9,21 +9,37 @@ import styles from './styles.scss';
 
 export interface DateOutputProps {
     className?: string;
-    value: string;
+    value: string | undefined;
+    label?: string;
+    tooltip?: string;
+    hideIcon?: boolean;
 }
 
 function DateOutput(props: DateOutputProps) {
     const {
         className,
         value,
+        label,
+        tooltip,
+        hideIcon,
     } = props;
 
     return (
-        <div className={_cs(styles.dateOutput, className)}>
-            <Icon
-                className={styles.icon}
-                name="calendar"
-            />
+        <div
+            className={_cs(styles.dateOutput, className)}
+            title={tooltip}
+        >
+            { ((value || label) && !hideIcon) && (
+                <Icon
+                    className={styles.icon}
+                    name="calendar"
+                />
+            )}
+            { label && (
+                <div className={styles.label}>
+                    { label }
+                </div>
+            )}
             <FormattedDate
                 className={styles.date}
                 value={value}
--- a/src/components/viewer/DateOutput/index.tsx
+++ b/src/components/viewer/DateOutput/index.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { _cs } from '@togglecorp/fujs';
+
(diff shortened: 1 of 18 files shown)


src/components/viewer/DateOutput/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { _cs } from '@togglecorp/fujs';
+
+import FormattedDate from '#rscv/FormattedDate';
+import Icon from '#rscg/Icon';
+
+import styles from './styles.scss';
+
+
+export interface DateOutputProps {
+    className?: string;
```
AdityaKhatri, line 11: Support title on hover
Also, add title show/hide behavior

src/views/Entries/requests/EntriesRequest.js
```diff
@@ -23,6 +23,8 @@ export default class EntriesRequest extends Request {
             count: totalEntriesCount,
         } = response;
 
+        console.info(responseEntries);
```
tnagorra: Let's not leave console.info intended for debugging.

src/constants/routes.js
```diff
@@ -167,7 +167,7 @@ export const routes = {
     editEntries: {
         order: 41,
         type: ROUTE.private,
-        path: '/projects/:projectId/leads/:leadId/entries/edit/',
+        path: '/projects/:projectId/leads/:leadId/entries/:entryId?/edit/',
```
AdityaKhatri: This is not necessary right?

src/views/Entries/QualityControl/index.tsx
```diff
@@ -29,7 +37,22 @@ function QualityControl(props: QualityControlProps) {
                 rightContainerClassName={styles.right}
                 rightChild={(
                     <div className={styles.entryList}>
-                        Entry list
+                        { leadGroupedEntriesList.map((leadWithEntries) => {
```
AdityaKhatri, line 40: Use list components

src/views/Entries/requests/EntriesRequest.js
AdityaKhatri: remove this

src/typings/entry.ts
```diff
@@ -0,0 +1,62 @@
+import { DatabaseEntityBase } from './common';
+
+type EntryType = 'excerpt' | 'image';
```
tnagorra: we should also consider another type 'dataSeries'

src/constants/routes.js
tnagorra: I think we should revert this if we are not using this!

src/components/viewer/DateOutput/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    value: string;
```
tnagorra: We should handle case where value is `undefined`

src/views/Entries/QualityControl/index.tsx
frozenhelium: This is only temporary and will be changed later

src/components/viewer/DateOutput/index.tsx
frozenhelium: done

PR #1154, opened 2020-10-06
Feature/entry verification qc card

frozenhelium:
Merge only after: https://github.com/the-deep/client/pull/1147

### DIFF
--- a/src/components/general/EntryVerify/index.tsx
+++ b/src/components/general/EntryVerify/index.tsx
@@ -33,20 +33,11 @@ interface ComponentProps {
     title?: string;
     entryId: number;
     leadId: number;
-    verified: boolean;
-    hide: (entryPermission: {}) => boolean;
+    value: boolean;
     className?: string;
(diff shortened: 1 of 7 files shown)


src/components/general/EntryVerify/index.tsx
```diff
@@ -108,22 +105,21 @@ const requestOptions: Requests<ComponentProps, Params> = {
...
-        setEntryVerificationPending,
+        onPendingChange,
     } = props;
 
     const {
         setEntryVerificationRequest,
     } = requests;
 
-    const [verified, setVerificationStatus] = useState(verifiedFromProps);
+    const [verified, setVerificationStatus] = useState(value);
```
tnagorra, line 119: Why use use a `verified` state when all it does is reflect on prop `value`

frozenhelium: This was already there, This component is very questionable and should be refactored. I wonder how it get merged.

tnagorra: Yes, looks very bad!

PR #1157, opened 2020-10-07
Fix reading of stale information on geoptions re-fetch

AdityaKhatri:


src/views/Entries/index.js
```diff
@@ -145,20 +149,163 @@ const leadKeySelector = d => d.id;
...
+
+            return ({
+                filters: [
+                    ...processedFilters,
+                    ...Object.entries(otherFilters),
+                ],
+            });
+        },
+        onMount: false,
+        onPropsChanged: ['entriesFilter', 'activePage'],
```
tnagorra, line 249: Shouldn't this also be called on onMount?

AdityaKhatri: No, we only fetch entries after geooptions and frameworks are loaded to avoid unintended data references.

tnagorra: There are various limitations with this apporach when we have entriesFilter/activePage as triggers.
And, the data loading is also slow.
We should rather focus on fixing the bad-data reference (we can just hide entries unless geo-options and frameworks are loaded)

src/views/Entries/index.js
```diff
@@ -307,19 +398,34 @@ export default class Entries extends React.PureComponent {
...
+        this.setState({
+            successGeoOptions: true,
+        }, this.startEntriesRequest);
+    }
+
+    handleFrameworkGetSuccess = () => {
+        this.setState({
+            successFramework: true,
+        }, this.startEntriesRequest);
+    }
```
tnagorra, line 427: Both of them starts entries request?

tnagorra: This looks very easy to break. We should change this behavior as well.

PR #1169, opened 2020-10-15
Add pagination and TOC Filter in quality control view

### LINKED ISSUE #1166: Entries QA: Fetch entries using pagination

### LINKED ISSUE #1167: Entries QA: Transform ToC filter to match with current entry matrix filters

samshara:
fixes #1166 
fixes #1167

styling required

### DIFF
--- a/src/components/TableOfContents/index.tsx
+++ b/src/components/TableOfContents/index.tsx
@@ -1,7 +1,7 @@
 import React, { useState, useCallback } from 'react';
 import { _cs, isDefined } from '@togglecorp/fujs';
 
-import PrimaryButton from '#rsca/Button/PrimaryButton';
+import Button from '#rsca/Button';
 import ListView from '#rsu/../v2/View/ListView';
 
 import styles from './styles.scss';
@@ -18,6 +18,7 @@ interface Props<T, K extends string | number>{
     value: {key: K; id: K } | undefined;
     level?: number;
     defaultCollapseLevel?: number;
+    className: string;
 }
 
 type ToCItemProps<T, K extends string | number > = Omit<Props<T, K>, 'options'> & {
@@ -62,52 +63,44 @@ function ToCItem<T, K extends string | number>(props: ToCItemProps<T, K>) {
     );
 
     const isSelected = id === value?.id;
-
-    if (children && children.length > 0) {
-        return (
-            <div>
-                <div className={styles.itemContainer}>
-                    <div
-                        className={_cs(
-                            styles.item,
-                            isSelected && styles.selected,
-                        )}
-                        onClick={handleClick}
-                        onKeyDown={noOp}
-                        role="button"
-                        tabIndex={0}
-                    >
-                        {title}
-                    </div>
-                    <PrimaryButton
-                        className={_cs(styles.expandButton)}
-                        onClick={handleCollapseToggle}
-                        transparent
-                        iconName={collapsed ? 'down' : 'up'}
-                    />
-                </div>
-                {!collapsed && (
-                    <TableOfContents
-                        {...props}
-                        options={children}
-                        level={level + 1}
-                    />
-                )}
-            </div>
-        );
-    }
+    const hasChildren = children && children.length > 0;
 
     return (
-        <div className={styles.itemContainer}>
-            <div
-                className={_cs(styles.item, isSelected && styles.selected)}
-                onClick={handleClick}
-                onKeyDown={noOp}
-                role="button"
-                tabIndex={-1}
-            >
-                {title}
+        <div className={_cs(
+            styles.tocItem,
+            isSelected && styles.active,
+            !collapsed && styles.expanded,
+        )}
+        >
+            <div className={styles.header}>
+                <div
+                    className={styles.heading}
+                    onClick={handleClick}
+                    onKeyDown={noOp}
+                    role="button"
+                    tabIndex={0}
+                >
+                    {title}
+                </div>
+                { hasChildren && (
+                    <div className={styles.actions}>
+                        <Button
+                            className={styles.expandButton}
+                            onClick={handleCollapseToggle}
+                            transparent
+                            iconName={collapsed ? 'chevronDown' : 'chevronUp'}
+                        />
+                    </div>
+                )}
             </div>
+            {hasChildren && !collapsed && (
+                <TableOfContents
+                    {...props}
+                    className={_cs(props.className, styles.children)}
+                    options={children}
+                    level={level + 1}
+                />
+            )}
         </div>
     );
 }
@@ -120,6 +113,7 @@ function TableOfContents<T, K extends string | number>(props: ToCListProps<T, K>
         options,
         idSelector,
         level = 0,
+        className,
         ...otherProps
     } = props;
 
@@ -132,7 +126,7 @@ function TableOfContents<T, K extends string | number>(props: ToCListProps<T, K>
 
     return (
         <ListView
-            className={styles.tocList}
+            className={_cs(className, styles.tableOfContents)}
             data={options}
             renderer={ToCItem}
             keySelector={idSelector}
--- a/src/components/TableOfContents/index.tsx
+++ b/src/components/TableOfContents/index.tsx
@@ -6,6 +6,8 @@ import ListView from '#rsu/../v2/View/ListView';
 
 import styles from './styles.scss';
 
+const noOp = () => {};
+
 interface Props<T, K extends string | number>{
     idSelector: (datum: T) => K;
     keySelector: (datum: T) => K | undefined;
@@ -64,16 +66,16 @@ function ToCItem<T, K extends string | number>(props: ToCItemProps<T, K>) {
     if (children && children.length > 0) {
         return (
             <div>
-                <div className={styles.container}>
+                <div className={styles.itemContainer}>
                     <div
                         className={_cs(
                             styles.item,
                             isSelected && styles.selected,
                         )}
                         onClick={handleClick}
-                        onKeyDown={handleClick}
+                        onKeyDown={noOp}
                         role="button"
-                        tabIndex={-1}
+                        tabIndex={0}
                     >
                         {title}
                     </div>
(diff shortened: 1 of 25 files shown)


samshara:
@frozenhelium the entry edit does not work when used with data pulled from the server. This might be due to the inconsistency of lead information present in each entry which slightly differs from that of the original lead information.

src/components/general/EntryVerify/index.tsx
```diff
@@ -117,16 +121,11 @@ function EntryVerify(props: Props) {
...
-    useEffect(() => {
-        setVerificationStatus(value);
-    }, [value]);
-
-    // FIXME: memoize this
-    const selectedOption = verificationStatusOptions.find(v => v.key === verified) ||
-                           verificationStatusOptions[1];
+    const selectedOption = useMemo(
+        () => verificationStatusOptions.find(v => v.key === value) ||
+            verificationStatusOptions[1],
```
tnagorra:
```suggestion
        () => verificationStatusOptions.find(v => v.key === value) ??
            verificationStatusOptions[1],
```

Also why not use `0`?

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -97,6 +96,12 @@ function EntryCard(props: EntryCardProps) {
         // TODO; disable all actions if pending
     }, []);
 
+    const [isVerified, setVerificationStatus] = React.useState<boolean>(entry.verified);
+
+    const handleEntryVerify = React.useCallback((status: boolean) => {
+        setVerificationStatus(status);
+    }, []);
```
tnagorra:
```suggestion
    const handleEntryVerify = setVerificationStatus;
```

src/views/Entries/QualityControl/index.tsx
```diff
@@ -78,29 +139,30 @@ function QualityControl(props: QualityControlProps) {
...
-                                    entry={e}
-                                    lead={lead}
-                                    framework={framework}
-                                    isDeleted={deletedEntries[e.id]}
-                                    onDelete={handleEntryDelete}
-                                />
-                            ));
-                        })}
+                        { pending && <LoadingAnimation /> }
+                        {response && response.results.map(e => (
```
AdityaKhatri: Use List component

src/components/TableOfContents/index.tsx
```diff
@@ -7,12 +7,13 @@ import ListView from '#rsu/../v2/View/ListView';
 import styles from './styles.scss';
 
 interface Props<T, K extends string | number>{
-    keySelector: (datum: T) => K;
+    idSelector: (datum: T) => K;
+    keySelector: (datum: T) => K | undefined;
```
tnagorra, line 13: What is the difference between key and id?

src/views/Entries/QualityControl/index.tsx
```diff
@@ -78,29 +139,30 @@ function QualityControl(props: QualityControlProps) {
...
+                                framework={framework}
+                                isDeleted={deletedEntries[e.id]}
+                                onDelete={handleEntryDelete}
+                            />
+                        ))}
                     </div>
                 )}
             />
+            <footer className={styles.footer}>
+                <Pager
```
AdityaKhatri, line 237: Lets use the same Pager for both pages but switch the data according to which page the user is in.
Also, for QC lets save the selected matrix header and active page / total entries count in redux.

src/components/TableOfContents/index.tsx
tnagorra: Do we even need key?

src/components/general/EntryVerify/index.tsx
samshara: 0 is for verified

src/components/TableOfContents/index.tsx
samshara: the server API requires it. I don't know why but it is as such right now.

src/components/general/EntryVerify/index.tsx
tnagorra: I think we should not hard-code this in this way. 
Thoughts @frozenhelium @AdityaKhatri ?

src/components/TableOfContents/index.tsx
tnagorra: Is key for client and id for server? Why do we need both?
Anyway, we need to document this through comments before we can merge this functionality.

AdityaKhatri: Each widget option has id and key which is prevelant throughout code, so i think its fine.
Still, I'll quickly add the reason behind having 2 things.

src/components/TableOfContents/index.tsx
```diff
@@ -29,17 +33,24 @@ function ToCItem<T, K extends string | number>(props: ToCItemProps<T, K>) {
...
     const children = childrenSelector(option);
 
     const handleClick = useCallback(
-        () => onChange(id),
-        [id, onChange],
+        () => {
+            if (isDefined(key)) {
+                return onChange({ key, id });
+            }
+            return undefined;
```
tnagorra: Why are we returning? onChange has no return value.

src/components/general/EntryVerify/index.tsx
```diff
@@ -117,16 +121,11 @@ function EntryVerify(props: Props) {
...
-    useEffect(() => {
-        setVerificationStatus(value);
-    }, [value]);
-
-    // FIXME: memoize this
-    const selectedOption = verificationStatusOptions.find(v => v.key === verified) ||
-                           verificationStatusOptions[1];
+    const selectedOption = useMemo(
+        () => verificationStatusOptions.find(v => v.key === value) ??
+            verificationStatusOptions[1],
```
tnagorra: We should think about a better way to identify a default value @thenav56 @frozenhelium @AdityaKhatri

tnagorra: At least, it shouldn't break silently.

src/redux/selectors/siloDomainData/entries.js
```diff
@@ -30,6 +30,21 @@ export const entriesViewActivePageSelector = createSelector(
     entriesView => entriesView.activePage || 1,
 );
 
+export const qualityControlViewActivePageSelector = createSelector(
+    entriesViewForProjectSelector,
+    entriesView => entriesView.qcViewActivePage || 1,
```
tnagorra: use `??`

src/redux/selectors/siloDomainData/entries.js
```diff
@@ -30,6 +30,21 @@ export const entriesViewActivePageSelector = createSelector(
...
+);
+
+export const qualityControlViewEntriesCountSelector = createSelector(
+    entriesViewForProjectSelector,
+    entriesView => entriesView.qcViewEntriesCount || 0,
```
tnagorra: Use `??`

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -91,12 +95,22 @@ function EntryCard(props: EntryCardProps) {
         isDeleted,
     } = props;
 
+    console.warn('here', lead);
```
tnagorra: console.warn on render meant for debuggin.

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -91,12 +95,22 @@ function EntryCard(props: EntryCardProps) {
...
+    const {
+        url: leadUrlFromProps,
+        attachment,
+    } = lead;
+
+    const leadUrl = (attachment && attachment.file) || leadUrlFromProps;
```
tnagorra: Use `??`

src/views/Entries/QualityControl/index.tsx
```diff
@@ -57,52 +113,132 @@ function QualityControl(props: QualityControlProps) {
...
+        body: {
+            filters: [
+                selected && Object.values(selected),
+                ...processedFilters,
+                ['project', projectId],
+            ].filter(isDefined),
+        },
+        method: 'POST',
+        onSuccess: (successResponse) => {
+            console.warn('here', successResponse);
```
tnagorra: console.warn meant for debugging

src/views/Entries/QualityControl/index.tsx
```diff
@@ -57,52 +113,132 @@ function QualityControl(props: QualityControlProps) {
...
+            setEntriesCount({ count: successResponse.count });
+        },
+    });
+
+    useEffect(
+        () => getEntries(),
```
tnagorra:
```suggestion
       getEntries,
```

src/views/Entries/QualityControl/index.tsx
```diff
@@ -57,52 +113,132 @@ function QualityControl(props: QualityControlProps) {
...
+    const handleEntryDelete = React.useCallback((entryId) => {
+        setDeletedEntries(oldDeletedEntries => ({ ...oldDeletedEntries, [entryId]: true }));
+    }, [setDeletedEntries]);
+
+    const handleSelection = useCallback((value) => {
+        if (selected && selected.id === value.id) {
+            setSelection({ matrixKey: undefined });
+        } else {
+            setSelection({ matrixKey: value });
+        }
```
tnagorra: We can use a ternary operator here

src/components/TableOfContents/index.tsx
samshara: some of TOC elements don't have the corresponding key and aren't filterable. undefined is returned as to be consistent with eslint rules.

tnagorra: The gist is that we don't need to return anything. Why return at all?

src/components/general/EntryVerify/index.tsx
AdityaKhatri: Not ideal, but works right now

src/typings/entry.ts
```diff
@@ -67,7 +69,12 @@ export interface EntryFields extends DatabaseEntityBase {
     image?: string;
     tabularField: number;
     tabularFieldData: TabularDataFields;
-    lead: number;
+    lead: Pick<Lead, EntryLeadType> & {
+        assingee: number[];
```
tnagorra:
```suggestion
        assignee: number[];
```

src/components/general/EntryVerify/index.tsx
tnagorra: "Works right now" doesn't work for me.

src/components/TableOfContents/index.tsx
tnagorra: Where was the reason added?

PR #1194, opened 2020-11-09
Add edit lead feature from EntryCard

### LINKED ISSUE #1177: Entries QA: Lead edit button
Add the ability to edit leads directly from the card

samshara:
fixes #1177
depends on https://github.com/the-deep/server/pull/563


src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -10,22 +10,31 @@ import EntryDeleteButton from '#components/general/EntryDeleteButton';
...
+import {
+    Permission,
+} from '#typings/common';
 import {
     FrameworkFields,
 } from '#typings/framework';
-
 import _ts from '#ts';
 
+import { LeadEditModal } from '../../../Leads/ActionButtons/LeadEditModal';
```
AdityaKhatri: Move this component to `components/general` and use this in both pages

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -90,32 +99,64 @@ function EntryCard(props: EntryCardProps) {
...
-    const leadSource = lead.sourceDetails ? lead.sourceDetails.title : lead.sourceRaw;
+    const leadSource = lead.sourceDetail ? lead.sourceDetail.title : lead.sourceRaw;
+
+    const [
+        pending,
+        leadFromRequest,
+        ,
+        getLead,
+    ] = useRequest<Lead>({
+        url: `server://v2/leads/${lead.id}`,
```
AdityaKhatri: end  the url with `/`

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -90,32 +99,64 @@ function EntryCard(props: EntryCardProps) {
...
+    const handleEditLeadButtonClick = () => {
+        getLead();
+    };
+
+    const handleEditLeadModalClose = () => {
+        showEditLeadModal(false);
+    };
+
+    const handleEntryEditSave = useCallback((lead: Lead) => {
+        setLead(lead);
```
AdityaKhatri: I think we should modify the data source instead of doing this only in this component?
Thoughts @samshara , @tnagorra ?

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -168,6 +209,27 @@ function EntryCard(props: EntryCardProps) {
                             }
                         />
                     )}
+                    <Cloak
+                        hide={shouldHideLeadEdit}
+                        render={
+                            <WarningButton
```
AdityaKhatri, line 217: Let's use modalize instead of handling it separately if possible.

AdityaKhatri: Is it because lead information is incomplete?

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -168,6 +209,27 @@ function EntryCard(props: EntryCardProps) {
...
+                                iconName="edit"
+                                transparent
+                                disabled={pending}
```
AdityaKhatri: Disable it with a popup is open as well.

src/views/Entries/QualityControl/EntryCard/index.tsx
samshara: patching the lead data can be done but I think it won't be useful here as when switching to the leads page the data is again pulled from the server. @tnagorra any thoughts?

src/views/Entries/QualityControl/EntryCard/index.tsx
samshara: it is because I had to send out the request on click

src/views/Entries/QualityControl/EntryCard/index.tsx
tnagorra: we shouldn't create a state for lead on individual card.
A change on lead inside a card should reflect that change to other entries card with the same lead

AdityaKhatri: @samshara , we might need to look for other entries with the same lead and change the data there as well.

AdityaKhatri (approved):

PR #1215, opened 2020-11-25
Show scale widgets information in entry card

### LINKED ISSUE #546: Entries QA: Show scale widgets` data at the end of entry card similar to word export

samshara:
fixes https://github.com/the-deep/server/issues/546

### DIFF
--- a/src/typings/entry.ts
+++ b/src/typings/entry.ts
@@ -1,3 +1,4 @@
+import { Attributes } from 'react';
 import { DatabaseEntityBase } from './common';
 import { Lead } from './lead';
 
@@ -16,9 +17,9 @@ export interface ProjectLabelFields {
 export interface AttributeFields {
     id: number;
     data?: {
-        value: {
+        value?: {
             [index: string]: unknown;
-        };
+        } | string;
     };
 }
 
@@ -56,7 +57,9 @@ export interface TabularDataFields {
 }
 
 export interface EntryFields extends DatabaseEntityBase {
-    attributes: AttributeFields[];
+    attributes: {
+        [key: string] : AttributeFields;
+    };
     analysisFramework: number;
     entryType: EntryType;
     project: number;
--- a/src/typings/entry.ts
+++ b/src/typings/entry.ts
@@ -1,4 +1,3 @@
-import { Attributes } from 'react';
 import { DatabaseEntityBase } from './common';
 import { Lead } from './lead';
 
--- a/src/typings/entry.ts
+++ b/src/typings/entry.ts
@@ -57,7 +57,7 @@ export interface TabularDataFields {
 
 export interface EntryFields extends DatabaseEntityBase {
     attributes: {
-        [key: string] : AttributeFields;
+        [key: string]: AttributeFields;
     };
     analysisFramework: number;
     entryType: EntryType;
--- a/src/utils/framework.ts
+++ b/src/utils/framework.ts
@@ -11,6 +11,10 @@ import {
     Matrix2dFlatSubdimensionElement,
     Matrix1dWidgetElement,
     MatrixTocElement,
+    ScaleWidget,
+    WidgetElement,
+    ConditionalWidget,
+    Entry,
 } from '#typings';
 
 export const SECTOR_FIRST = 'sectorFirst';
@@ -73,6 +77,20 @@ interface LevelOutput {
     selected: boolean;
     draggable: boolean;
     nodes?: LevelOutput[];
+}
+
+interface WidgetData {
+    data: {
+        value: string;
+    };
+}
+interface ConditionalAttribute {
+    selectedWidgetKey?: string;
+    [key: string]: WidgetData | string | undefined;
+}
+
+function isWidgetData(arg: any): arg is WidgetData {
+    return arg?.data?.value !== undefined;
 }
 
 export function mapReportLevelsToNodes(levels: Level[]): LevelOutput[] {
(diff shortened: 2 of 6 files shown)


src/utils/framework.ts
```diff
@@ -399,3 +416,69 @@ export const getMatrix2dToc = (framework: FrameworkFields | undefined): MatrixTo
...
+                                            scaleUnits,
+                                        },
+                                    },
+                                } = widget as WidgetElement<ScaleWidget>;
+
+                                return scaleUnits.find(v => v.key === attributeData.data.value);
+                            }
+                        }
+                    }
+                    return undefined;
```
AdityaKhatri: Return undefined if selectedWidgetKey is not defined immediately, will reduce a level of nesting

src/utils/framework.ts
```diff
@@ -75,6 +79,20 @@ interface LevelOutput {
...
+    };
+}
+interface ConditionalAttribute {
+    selectedWidgetKey?: string;
+    [key: string]: WidgetData | string | undefined;
+}
+
+function isWidgetData(arg: any): arg is WidgetData {
+    return arg?.data?.value !== undefined;
+}
```
tnagorra, line 94: This doesn't look right

src/typings/entry.ts
```diff
@@ -16,9 +16,9 @@ export interface ProjectLabelFields {
 export interface AttributeFields {
     id: number;
     data?: {
-        value: {
+        value?: {
             [index: string]: unknown;
-        };
+        } | string;
```
tnagorra, line 21: This doesn't look right. When is the value object vs string?

src/utils/framework.ts
```diff
@@ -75,6 +79,20 @@ interface LevelOutput {
     nodes?: LevelOutput[];
 }
 
+interface WidgetData {
+    data: {
+        value: string;
```
tnagorra, line 84: Is value always string for WidgetData?

src/utils/framework.ts
```diff
@@ -75,6 +79,20 @@ interface LevelOutput {
...
+    };
+}
+interface ConditionalAttribute {
+    selectedWidgetKey?: string;
+    [key: string]: WidgetData | string | undefined;
```
tnagorra, line 89: Do we have both WidgetData and string on same mapping?

src/utils/framework.ts
```diff
@@ -399,3 +416,67 @@ export const getMatrix2dToc = (framework: FrameworkFields | undefined): MatrixTo
...
+
+export function getScaleWidgetsData(framework: FrameworkFields, entry: Entry) {
+    if (!framework || !framework.widgets) {
+        return [];
+    }
+
+    const { widgets } = framework;
+
+    const scaleWidgets = widgets
+        .filter(w => w.widgetId === 'scaleWidget')
```
tnagorra, line 428: We should use widgetId on widgets to identify if it is ScaleWidget, we wil no longer need to cast it ourself

src/utils/framework.ts
```diff
@@ -399,3 +416,67 @@ export const getMatrix2dToc = (framework: FrameworkFields | undefined): MatrixTo
...
+    const scaleWidgetsInsideConditionals = widgets
+        .filter(w => w.widgetId === 'conditionalWidget')
+        .map((conditional) => {
+            const {
+                id,
+                properties: {
+                    data: {
+                        widgets: widgetsInsideConditional = [],
+                    } = {},
+                } = {},
```
tnagorra: In typescript, it's better to just use optional chaining instead of setting default values

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -175,9 +186,19 @@ function EntryCard(props: EntryCardProps) {
 
     const loading = verifiyChangePending;
 
+    const scaleWidgets = useMemo(() => getScaleWidgetsData(framework, entry), [framework, entry]);
+
+    const scaleWidgetRendererParams = (_: string, d: ScaleWidget) => {
```
tnagorra: use useCallback

src/views/Entries/QualityControl/EntryCard/index.tsx
```diff
@@ -212,6 +233,15 @@ function EntryCard(props: EntryCardProps) {
                             </div>
                         )}
                     </div>
+                    <div>
```
tnagorra: Un-necessary nested div

src/typings/entry.ts
samshara: for conditional widgets attributes, it's an object and rests otherwise

src/utils/framework.ts
samshara: I made the interface based on the incoming data. I don't exactly know whether all values are string or other values are possible too.

src/utils/framework.ts
samshara: i added string value as to comply selectedWidgetKey property with the second property specification

src/utils/framework.ts
AdityaKhatri: @tnagorra what do you mean?

src/utils/framework.ts
tnagorra: Maybe it differs from widget to widget

src/utils/framework.ts
tnagorra: No worries!

src/utils/framework.ts
tnagorra: What the filter does is act as a type guard, so if that widget has a string literal as identifier for different unions we wont' need to cast explicitly.

PR #1216, opened 2020-11-26
Refactor user exports

AdityaKhatri:
Depends on: https://github.com/the-deep/server/pull/576


src/views/UserExports/index.js
```diff
@@ -1,383 +1,313 @@
...
+            ordering: activeSort,
+            is_preview: false,
+            offset: (activePage - 1) * maxItemsPerPage,
+            limit: maxItemsPerPage,
+        },
+        method: 'GET',
+        autoTrigger: true,
+        shouldPoll: (response) => {
+            const isPending = response?.results.some(e => e.status === 'pending' || e.status === 'started');
+            if (isPending) {
```
samshara, line 76: why

src/views/UserExports/index.js
```diff
@@ -1,383 +1,313 @@
...
+                return 5000;
+            }
```
samshara, line 78: why return 5000?

AdityaKhatri: Power level more than 5000!!
Its the polling time

PR #1234, opened 2020-12-08
Fix/connector trigger status

AdityaKhatri:


src/views/Project/Details/Connector/ConnectorEditForm/index.tsx
```diff
@@ -369,6 +388,12 @@ function ProjectConnectorEditForm(props: OwnProps) {
                     </FaramList>
                 </ModalBody>
                 <ModalFooter className={styles.modalFooter}>
+                    <Checkbox
+                        value={autoTrigger}
+                        onChange={setAutoTrigger}
+                        label={_ts('project.connector.editForm', 'triggerLeadsExtraction')}
+                        disabled={pending}
+                    />
```
samshara, line 399: shouldn't this change pristine value so that it could be the form could be saved without changing other parameters.

AdityaKhatri: This is not a part of the form so not worries

AdityaKhatri: Also, to retrigger there is a button outside as well, so they'll do that if they need to retrigger the extraction

samshara (approved):

PR #1237, opened 2020-12-09
Bump eslint to support new typescript syntax and fix linting errors

samshara:


src/components/TableOfContents/index.tsx
```diff
@@ -14,7 +14,7 @@ import { hasKey } from '#utils/common';
 
 import styles from './styles.scss';
 
-const noOp = () => {};
+const noOp = () => ({});
```
tnagorra: This doesn't look correct. noop should return nothing.
This change will make noop return an empty object;

src/components/TableOfContents/index.tsx
```diff
@@ -104,7 +103,11 @@ function ToCItem<T, K extends string | number>(props: ToCItemProps<T, K>) {
 
     useEffect(() => {
         const isCollapsed = isDefined(defaultCollapseLevel) && level >= defaultCollapseLevel;
-        isCollapsed ? setCollapsed(true) : setCollapsed(false);
+        if (isCollapsed) {
+            setCollapsed(true);
+        } else {
+            setCollapsed(false);
+        }
```
tnagorra:
```suggestion
        setCollapsed(isCollapsed);
```

PR #1242, opened 2020-12-10
Fix non unique id for matrix1D and matrix2D elements in ToC

samshara:


src/views/Entries/QualityControl/index.tsx
```diff
@@ -345,6 +337,10 @@ function QualityControl(props: Props) {
         }
     }, [defaultCollapseLevel]);
 
+    const handleSearch = useCallback((value) => {
+        setSearchValue(value);
+    }, []);
```
AdityaKhatri: use setSearchValue directly instead of this callback

PR #1251, opened 2020-12-15
Add authoring organization type in leads filter

subinasr:


src/components/other/FilterLeadsForm/index.js
```diff
@@ -240,6 +241,7 @@ export default class FilterLeadsForm extends React.PureComponent {
                 status,
                 assignee,
                 priority,
+                organizationTypes,
```
AdityaKhatri, line 244: add this to schema

PR #1252, opened 2020-12-15
Add notify error on project creation form

AdityaKhatri:


src/components/other/ProjectAddForm/index.js
```diff
@@ -58,142 +60,131 @@ const mapDispatchToProps = dispatch => ({
...
+        onSuccess: (response) => {
+            setUserProject({ project: response });
+            setUsergroupProject({ project: response });
+            setUserProfileProject({
+                userId,
+                project: response,
+            });
+            if (onProjectAdd) {
+                onProjectAdd(response.id);
+            }
```
tnagorra, line 107: Is this correct? Shouldn't onProjectAdd accept `response` then the parent handle setUserProject, setUsergroupProject and setUserProfileProject on the parent itself?

src/components/other/ProjectAddForm/index.js
```diff
@@ -58,142 +60,131 @@ const mapDispatchToProps = dispatch => ({
...
+                message: _ts('components.addProject', 'userProjectCreateSuccess'),
+                duration: notify.duration.MEDIUM,
+            });
+            onModalClose();
+        },
+        onFailure: (error, errorBody) => {
+            setFaramErrors(errorBody?.faramErrors);
+            setPristine(true);
+            notifyOnFailure(_ts('components.addProject', 'addProjectModalLabel'))({ error: errorBody });
+        },
```
tnagorra, line 119: handle Fatal condition

src/components/other/ProjectAddForm/index.js
```diff
@@ -58,142 +60,131 @@ const mapDispatchToProps = dispatch => ({
...
+
+    const handleFaramChange = useCallback((newFaramValues, newFaramErrors) => {
+        setFaramValues(newFaramValues);
+        setFaramErrors(newFaramErrors);
+        setPristine(false);
+    }, [setFaramValues, setFaramErrors, setPristine]);
+
+    const handleFaramFailure = useCallback((newFaramErrors) => {
+        setFaramErrors(newFaramErrors);
+        setPristine(true);
```
tnagorra: failure shouid not set pristine to true

src/components/other/ProjectAddForm/index.js
```diff
@@ -58,142 +60,131 @@ const mapDispatchToProps = dispatch => ({
...
+    }, [setFaramErrors, setPristine]);
+
+    const handleFaramSuccess = useCallback(() => {
+        createNewProject();
+    }, [createNewProject]);
```
tnagorra: const handleFaramSuccess = createNewProject;

samshara: if it is so then can't we use createNewProject directly as props to Faram.

src/components/other/ProjectAddForm/index.js
AdityaKhatri: This was done previously and I replicated it

src/components/other/ProjectAddForm/index.js
AdityaKhatri: I was under the impression that the new request doesn't have fatal

src/components/other/ProjectAddForm/index.js
AdityaKhatri: yes it should, why not?

samshara (approved):

src/components/other/ProjectAddForm/index.js
tnagorra: we can yes sir!

src/components/other/ProjectAddForm/index.js
AdityaKhatri: I've added a note

PR #1259, opened 2020-12-18
Add sub-sector selection in report structure of export page

### LINKED ISSUE #1225: Export: Add 2 options in report structure to support sub-sectors
Support Row>Column and Column>Row with sub-sectors

samshara:
fixes https://github.com/the-deep/client/issues/1225
depends on https://github.com/the-deep/server/pull/620

### DIFF
--- a/src/utils/framework.ts
+++ b/src/utils/framework.ts
@@ -22,35 +22,37 @@ import {
 export const SECTOR_FIRST = 'sectorFirst' as const;
 export const DIMENSION_FIRST = 'dimensionFirst' as const;
 
+function createId<T extends string | number>(...args: [T, T, ...T[]]) {
+    return args.join('-');
+}
+
 // NOTE: This function generates dimension first level
-const transformLevelsDimensionFirst = ({
-    dimensions: widgetDim,
-    sectors: widgetSec,
-}: Matrix2dWidgetElement['properties']['data']) => {
+const transformLevelsDimensionFirst = (
+    {
+        dimensions: widgetDim,
+        sectors: widgetSec,
+    }: Matrix2dWidgetElement['properties']['data'],
+    includeSubSector: boolean,
+) => {
     const dimensionFirstLevels = widgetDim.map((d) => {
         const subDims = d.subdimensions;
 
         const sublevels = subDims.map((sd) => {
             const sectors = widgetSec.map((s) => {
                 const { subsectors } = s;
-                if (subsectors) {
-                    const subSectorLevels = subsectors.map(ss => ({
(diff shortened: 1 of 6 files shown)


AdityaKhatri (approved):

src/utils/framework.ts
```diff
@@ -22,34 +22,36 @@ import {
...
+                    return ({
+                        id: `${s.id}-${d.id}-${sd.id}`,
+                        title: s.title,
+                        sublevels: subSectorLevels,
+                    });
+                }
+                return ({
+                    id: `${s.id}-${d.id}-${sd.id}`,
+                    title: s.title,
+                });
```
tnagorra:
```suggestion
                return ({
                    id: `${s.id}-${d.id}-${sd.id}`,
                    title: s.title,
                    sublevels: includeSubSector
                        ? subsectors?.map(ss => ({
                            id: `${s.id}-${ss.id}-${d.id}-${sd.id}`,
                            title: ss.title,
                        }))
                        : undefined,
                });
```

tnagorra: It's better to do things once

tnagorra: In this case, setting the id for sector.

src/utils/framework.ts
```diff
@@ -67,6 +69,58 @@ const transformMatrix2dLevels = ({
...
+                    id: `${s.id}-${ss.id}`,
+                    title: ss.title,
+                    sublevels: dimensions,
+                });
+            });
+        } else {
+            sublevels = widgetDim.map((d) => {
+                const { subdimensions } = d;
+                const subDimensionsLevel = subdimensions.map(sd => ({
+                    id: `${s.id}-${d.id}-${sd.id}`,
```
tnagorra: Maybe we should create a simple function to do this repeated logic

```typescript
function createId<T extends string | number>(...args: [T, T, ...T[]]) {
    return args.join('-');
}
```

tnagorra: This function will not accept null, undefined, or arguments with length less than 2.

src/views/Export/index.tsx
```diff
@@ -234,9 +235,13 @@ function Export(props: Props) {
...
     useEffect(() => {
-        const structure = createReportStructure(analysisFramework, reportStructureVariant);
+        const structure = createReportStructure(
+            analysisFramework,
+            reportStructureVariant,
+            includeSubSector,
+        );
         setReportStructure(structure);
-    }, [analysisFramework, reportStructureVariant]);
+    }, [analysisFramework, reportStructureVariant, includeSubSector]);
```
tnagorra, line 244: Can't we just use a useMemo for `reportStructure` value?

PR #1260, opened 2020-12-21
Feature/entry verify edit page

AdityaKhatri:
Depends on: https://github.com/the-deep/server/pull/619

### DIFF
--- a/src/redux/reducers/siloDomainData/editEntries.js
+++ b/src/redux/reducers/siloDomainData/editEntries.js
@@ -37,6 +37,8 @@ export const EEB__CLEAR_ENTRIES = 'siloDomainData/EEB__CLEAR_ENTRIES';
 
 export const EEB__SET_ENTRIES_COMMENTS_COUNT = 'siloDomainData/EEB__SET_ENTRIES_COMMENTS_COUNT';
 export const EEB__SET_ENTRY_COMMENTS_COUNT = 'siloDomainData/EEB__SET_ENTRY_COMMENTS_COUNT';
+export const EEB__SET_ENTRIES_VERIFICATION_STATUS = 'siloDomainData/EEB__SET_ENTRIES_VERIFICATION_STATUS';
+export const EEB__SET_ENTRY_VERIFICATION_STATUS = 'siloDomainData/EEB__SET_ENTRY_VERIFICATION_STATUS';
 
 export const EEB__SET_SELECTED_ENTRY_KEY = 'siloDomainData/EEB__SET_SELECTED_ENTRY_KEY';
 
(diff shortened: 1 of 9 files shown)


samshara (approved):

src/redux/reducers/siloDomainData/editEntries.js
```diff
@@ -420,6 +434,89 @@ const setEntryCommentsCount = (state, action) => {
...
+        entries: entriesFromServer,
+    } = action;
+
+    const {
+        editEntries: {
+            [leadId]: {
+                entries = [],
+            } = {},
+        } = {},
+    } = state;
```
tnagorra: For these kind of access, we can just use optional chaining now.


```suggestion
const entries = state.?editEntries.?[leadId].?entries ?? [];
```

src/redux/reducers/siloDomainData/editEntries.js
```diff
@@ -420,6 +434,89 @@ const setEntryCommentsCount = (state, action) => {
...
+        }
+        if (!safeState.editEntries[leadId].entries) {
+            // eslint-disable-next-line no-param-reassign
+            safeState.editEntries[leadId].entries = [];
+        }
+        const index = entries.findIndex(e => entry.entryId === entryAccessor.serverId(e));
+        if (index > -1) {
+            // eslint-disable-next-line no-param-reassign
+            safeState.editEntries[leadId].entries[index]
+                .serverData.versionId = entry.versionId;
```
tnagorra: I think it's best to not update the versionId. It will only silently increase  the version without updating the real value associated with that version.

src/redux/reducers/siloDomainData/editEntries.js
```diff
@@ -420,6 +434,89 @@ const setEntryCommentsCount = (state, action) => {
...
+
+    const {
+        editEntries: {
+            [leadId]: {
+                entries = [],
+            } = {},
+        } = {},
+    } = state;
+
+    const newState = produce(state, (safeState) => {
```
tnagorra: We can just call `setEntriesVerificationStatus` here instead of writing two different reducers

src/views/EditEntries/List/WidgetFaramContainer/index.js
```diff
@@ -72,37 +75,56 @@ const mapStateToProps = state => ({
...
-
-    containerRef = React.createRef();
-    dragEnterCount = 0;
-
-    shouldHideEntryDelete = ({ entryPermissions }) => (
-        !entryPermissions.delete && !!entryAccessor.serverId(this.props.entry)
-    )
+function WidgetFaramContainer(props) {
+    const {
+        widgets, // eslint-disable-line no-unused-vars
```
tnagorra: Isn't widgets used? this looks odd

src/views/EditEntries/List/WidgetFaramContainer/index.js
```diff
@@ -128,169 +144,185 @@ export default class WidgetFaramContainer extends React.PureComponent {
...
+                leadId={leadId}
+            />
+        </div>
+    );
 }
+
+WidgetFaramContainer.propTypes = propTypes;
+WidgetFaramContainer.defaultProps = defaultProps;
+
+export default connect(mapStateToProps, mapDispatchToProps)(WidgetFaramContainer);
```
tnagorra: You may need to use memo

src/redux/reducers/siloDomainData/editEntries.js
AdityaKhatri: I think its fine as we reject the verification if the version is mismatched.

tnagorra: Validation is a metadata so it should not be concerned with the version id.

We should send the current version id to the server (or maybe +1) and restrict users to send validation if they have an older version.
If it does that, we can increase the versionId (but i still prefer if it doesn't). And, update that to the client side.
We can do this only if we have that restriction.

PR #1370, opened 2021-02-04
Feature/dui container

frozenhelium:


src/components/ui/Container/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface Props {
+    className?: string;
+    heading?: React.ReactNode;
+    headerIcons?: React.ReactNode;
+    headerActions?: React.ReactNode;
+    children?: React.ReactNode;
+    contentClassName?: string;
+    footerContent?: React.ReactNode;
+    footerActions?: React.ReactNode;
+    sub?: boolean;
```
AdityaKhatri, line 20: Add note on what `sub` is

AdityaKhatri:
For `Link` component, we should support external links as well, with a `type` prop in necessary.

AdityaKhatri (commented):
For `Link` component, we should support external links as well, with a `type` prop in necessary.

PR #1371, opened 2021-02-04
Add Tag component

samshara:


AdityaKhatri (changes requested):
Lets add at least one examples of:
1. Using icons
2. Using actions
3. Using both icons and actions

frozenhelium:
@AdityaKhatri @samshara I've fixed the usage and the styling method, please take a look at the diffs for the future reference

PR #1410, opened 2021-03-05
Add entry quick edit option

AdityaKhatri:


src/redux/reducers/siloDomainData/entries.js
```diff
@@ -261,6 +269,63 @@ const deleteEntry = (state, action) => {
...
+            const entryIndex = safeLeads[leadIndex].entries.findIndex(e => entryId === e.id);
+            const safeEntries = safeLeads[leadIndex].entries;
+
+            if (entryIndex > -1) {
+                // eslint-disable-next-line no-param-reassign
+                safeEntries.splice(entryIndex, 1);
+                safeEntries[entryIndex] = {
+                    ...entry,
+                    lead: safeEntries[entryIndex].lead,
+                };
```
tnagorra: What is the intention here?

samshara (approved):

src/redux/reducers/siloDomainData/entries.js
AdityaKhatri: The lead data is not changed while editing, but we fetch leads with a bit more detail.
So, we are just merging the edited entry data with old lead data.

tnagorra (approved):

PR #1414, opened 2021-03-09
Feature/timeline

frozenhelium:

### DIFF
--- a/src/components/viz/Timeline/index.tsx
+++ b/src/components/viz/Timeline/index.tsx
@@ -3,18 +3,103 @@ import { _cs } from '@togglecorp/fujs';
 
 import styles from './styles.scss';
 
-export interface TimelineProps {
+export type MinMax = [number, number];
+
+export interface TimelineProps<T> {
     className?: string;
+    data: T[];
+    keySelector: (item: T) => string | number;
+    valueSelector: (item: T) => number;
+    labelSelector: (item: T) => React.ReactNode;
+    tickLabelSelector: (item: number) => React.ReactNode;
+    domain?: MinMax;
 }
 
-function Timeline(props: TimelineProps) {
+
+const scale = (domain: MinMax, range: MinMax, value: number) => {
+    const [minDomain, maxDomain] = domain;
+    const [minRange, maxRange] = range;
+    const normalizedValue = (value - minDomain) / (maxDomain - minDomain);
+    const newValue = (normalizedValue * (maxRange - minRange)) + minRange;
+    return newValue;
+};
+
+function Timeline<T>(props: TimelineProps<T>) {
     const {
         className,
+        labelSelector,
+        valueSelector,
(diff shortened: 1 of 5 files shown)


src/components/viz/Timeline/index.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
```
tnagorra: Should we not move this to #dui?

src/components/viz/Timeline/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        valueSelector,
+        keySelector,
+        tickLabelSelector,
+        data,
+        domain: domainFromProps,
+    } = props;
+
+    const domain: MinMax = React.useMemo(() => domainFromProps ?? [
+        Math.min(...data.map(valueSelector)),
+        Math.max(...data.map(valueSelector)),
```
tnagorra: We should verify the length of data here.
Math.min() is Infinity and Math.max() is -Infinity.

src/components/viz/Timeline/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        const [minDomain, maxDomain] = domain;
+        const maxTicks = 5;
+
+        const domainLength = maxDomain - minDomain;
+        const increment = domainLength / (maxTicks - 1);
+
+        const ticks = [];
+        for (let i = 0; i < maxTicks; i += 1) {
+            ticks.push(Math.round(minDomain + (i * increment)));
+        }
```
tnagorra:
```suggestion
        const ticks = Array.from({ length: maxTicks }, (v, i) => Math.round(minDomain + (i * increment)));
```

src/components/viz/Timeline/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+        if (domainFromProps) {
+            return domainFromProps;
+        }
+        if (data.length > 0) {
+            return ([
+                Math.min(...data.map(valueSelector)),
+                Math.max(...data.map(valueSelector)),
+            ]);
+        }
+        return ([0, 0]);
```
tnagorra:
```suggestion
        return [0, 0];
```

src/components/viz/Timeline/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+    const domain: MinMax = React.useMemo(() => {
+        if (domainFromProps) {
+            return domainFromProps;
+        }
+        if (data.length > 0) {
+            return ([
+                Math.min(...data.map(valueSelector)),
+                Math.max(...data.map(valueSelector)),
+            ]);
```
tnagorra:
```suggestion
            return [
                Math.min(...data.map(valueSelector)),
                Math.max(...data.map(valueSelector)),
            ];
```

PR #1421, opened 2021-03-12
Add toggle entry control button

frozenhelium:


samshara (approved):

src/utils/request/index.ts
```diff
@@ -130,7 +130,7 @@ async function fetchResource<T>(
                         faramErrors
                         && faramErrors.$internal
                         && faramErrors.$internal.join(' ')
-                    ) || 'There was some error while performing this action. Please try again.';
+                    ) || 'There was some error while performing this action';
```
AdityaKhatri, line 133: There were*

frozenhelium: Also use translation

PR #1472, opened 2021-03-24
Add entries viz share URL

AdityaKhatri:
Depends on: https://github.com/the-deep/server/pull/702

### DIFF
--- a/src/components/other/ProjectAddForm/index.js
+++ b/src/components/other/ProjectAddForm/index.js
@@ -68,6 +68,8 @@ const projectAddSchema = {
     },
 };
 
+const shouldHidePrivacyChangeButton = ({ accessPrivateProject }) => !accessPrivateProject;
+
 function ProjectAddForm(props) {
     const {
         className,
@@ -157,7 +159,7 @@ function ProjectAddForm(props) {
                 autoFocus
             />
             <Cloak
-                hide={ProjectAddForm.shouldHidePrivate}
+                hide={shouldHidePrivacyChangeButton}
                 render={
                     <SegmentInput
                         options={projectVisibilityOptions}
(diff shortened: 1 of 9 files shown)


src/components/other/ProjectAddForm/index.js
```diff
@@ -68,6 +68,8 @@ const projectAddSchema = {
     },
 };
 
+const shouldHidePrivate = ({ accessPrivateProject }) => !accessPrivateProject;
```
tnagorra:
```suggestion
const shouldHidePrivateProject = ({ accessPrivateProject }) => !accessPrivateProject;
```

src/redux/initial-state/dev-lang.json
```diff
@@ -1299,7 +1299,12 @@
         "-98549020": "Select Analysis Date",
         "-22196585": "Create a project",
         "-48980925": "Save and get started",
-        "-59904873": "Project details"
+        "-59904873": "Project details",
+        "-99774780": "The generated link can be publicly accessed by anyone without authentication and contains only non-confidential data.",
+        "-79488121": "Share",
+        "-99801867": "Share public link",
+        "-3100650": "Share public link",
```
tnagorra: But why do we have different entries for same text?

tnagorra (approved):

tnagorra (changes requested):
Need some fixes!

src/views/Entries/ShareModal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    onShareLinkChange(response?.publicUrl);
+                }
+            },
+            onFailure: (_, errorBody) =>
+                notifyOnFailure(_ts('entries', 'shareVizLink'))({ error: errorBody }),
+        },
+    );
+
+    return (
+        <Modal
```
subinasr, line 76: Shouldn't we be using the new Modal component?

subinasr (approved):

tnagorra (approved):

src/views/Entries/ShareModal/index.tsx
tnagorra: The new UI is not used on the Entries page. So, no!

PR #1589, opened 2021-05-05
Restrict lead attachement fetch on lead edit modal

AdityaKhatri:


src/views/LeadAdd/LeadDetail/index.js
```diff
@@ -424,12 +424,13 @@ class LeadDetail extends React.PureComponent {
...
             requests: { fileUrlGetRequest },
             lead,
+            onLeadAttachmentChange,
         } = this.props;
 
         const values = leadFaramValuesSelector(lead);
         const { attachment } = values;
 
-        if (isDefined(attachment)) {
+        if (isDefined(onLeadAttachmentChange) && isDefined(attachment)) {
```
tnagorra: This is not a proper way to handle this behavior.
This doesn't not answer why we are doing this.
We need a prop to enable/disable this behavior (eg: shouldFetchAttachment)
We should also check this prop on fileUrlGetRequest success

tnagorra (approved):

src/views/LeadAdd/LeadDetail/index.js
```diff
@@ -424,12 +427,13 @@ class LeadDetail extends React.PureComponent {
         const {
```
tnagorra: Let's move this componentDidMount logic to onMount of fileUrlGetRequest

tnagorra (approved):

PR #1593, opened 2021-05-06
Fix entries viz share button visibility

AdityaKhatri:


src/views/Entries/ShareModal/index.tsx
```diff
@@ -30,7 +27,8 @@ interface ComponentProps {
     publicUrl?: string;
     projectId?: number;
     closeModal?: () => void;
-    onShareLinkChange?: (publicUrl?: string) => void;
+    onShareLinkChange?: (publicUrl?: string, publicShare?: boolean) => void;
```
tnagorra:
```suggestion
    onShareLinkChange?: (publicUrl: string | undefined, publicShare: boolean | undefined) => void;
```

tnagorra (changes requested):

src/views/Entries/ShareModal/index.tsx
```diff
@@ -48,11 +46,12 @@ function ShareModal(props: ComponentProps) {
         className,
         onShareLinkChange,
         publicUrl,
+        isEntriesVizPublic,
         closeModal,
         projectId,
     } = props;
 
-    const [publicUrlAvailable, setPublicUrlAvailability] = useState(isDefined(publicUrl));
+    const [publicUrlAvailable, setPublicUrlAvailability] = useState(isEntriesVizPublic);
```
tnagorra: Why not move this state outside this component altogether?

tnagorra (approved):

PR #1597, opened 2021-05-09
Patch entry control status change

samshara:
This commit updates control status for both EditEntries page and
Entries page. Previously the control status were updated only in
the respective view store.
depends on https://github.com/the-deep/server/pull/755

### DIFF
--- a/src/entities/editEntries.js
+++ b/src/entities/editEntries.js
@@ -161,7 +161,7 @@ export const createEntry = ({
     key,
     serverId,
     versionId,
-    verified,
+    controlled,
     data = {},
     isPristine = false,
     hasError = false,
@@ -204,8 +204,8 @@ export const createEntry = ({
             isMarkedAsDeleted: false,
         },
         serverData: {
+            controlled: false,
             versionId: undefined,
-            verified: false,
         },
         data: {
             id: undefined, // serverId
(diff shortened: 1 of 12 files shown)


src/entities/editEntries.js
```diff
@@ -204,6 +205,7 @@ export const createEntry = ({
             isMarkedAsDeleted: false,
         },
         serverData: {
+            controlled: false,
             versionId: undefined,
             verified: false,
```
AdityaKhatri: We might need to remove the old verified status

src/entities/editEntries.js
```diff
@@ -221,6 +223,7 @@ export const createEntry = ({
             error: { $set: undefined },
         },
         serverData: {
+            controlled: { $set: controlled },
             versionId: { $set: versionId },
             verified: { $set: verified },
```
AdityaKhatri: Remove the old verified status

src/schema/entries.js
```diff
@@ -49,7 +49,7 @@ const entrySchema = [];
         fields: {
             analysisFramework: { type: 'uint', required: true },
 
-            image: { type: 'number' },
+            image: { type: 'string' },
```
AdityaKhatri: This is number after you pull latest server

AdityaKhatri (changes requested):
Fix change requests

PR #1617, opened 2021-05-13
Add FrameworkDetails tab to view and edit basic framework information

samshara:


src/views/AnalyticalFramework/FrameworkDetails/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const {
+        pending: frameworkGetPending,
+        response: analyticalFramework,
+    } = useRequest<AnalyticalFramework>({
+        skip: isNotDefined(frameworkId),
+        url: `server://analysis-frameworks/${frameworkId}/`,
+        method: 'GET',
+        onFailure: (_, errorBody) =>
+            notifyOnFailure(_ts('analyticalFramework', 'title'))({ error: errorBody }),
+    });
```
AdityaKhatri, line 31: Lets fetch this in AnalyticalFramework view as its values will be used in other tabs as well.
If we do that, we can directly place the Form component outside and we can remove one level of nesting.

AdityaKhatri (changes requested):
Fix review requests

src/views/AnalyticalFramework/FrameworkDetails/index.tsx
samshara: Let's do this when it is necessary. We need to add Framework's users table too in a later phase so the nesting of Form component makes sense here.

src/views/AnalyticalFramework/FrameworkDetails/FrameworkDetailsForm/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+                            className={styles.createdOn}
+                            name="createdAt"
+                            value={value.createdAt?.split('T')[0]}
+                            readOnly
+                            label={_ts('analyticalFramework', 'createdOn')}
+                        />
+                    </div>
+                    <SelectInput
+                        name="organization"
+                        options={[]}
```
tnagorra: Is this a work in progress?

src/views/AnalyticalFramework/FrameworkDetails/FrameworkDetailsForm/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+        onValueChange,
+        validate,
+        onValueSet,
+        onErrorSet,
+    } = useForm(analyticalFrameworkFromProps ?? defaultFormValues, schema);
+
+    const {
+        pending: frameworkPatchPending,
+        trigger: patchFramework,
+    } = useLazyRequest<AnalyticalFramework, unknown>({
```
tnagorra: The second type should not be unknown

tnagorra (changes requested):

PR #1657, opened 2021-05-14
Show analytical statements and entries count  on pillar

tnagorra:
- Show actual stats on overview


src/views/AnalysisModule/Analysis/AnalysisPillar/index.tsx
```diff
@@ -127,6 +129,26 @@ function AnalysisPillar(props: ComponentProps) {
                             noColon
                         />
                     </div>
+                    <div className={styles.statementsItem}>
+                        <div>
+                            Statements
```
AdityaKhatri: use strings

PR #1669, opened 2021-05-21
Refactor AnalysisModule to fetch list after edit/add

samshara:
** Refetch the list after the analysis module is edited or added.
** Add validation for having at least one pillar analysis.
** Remove local cache


AdityaKhatri:
@samshara , we might have missed it during defining the AC, but right now there are multiple requests per analysis being sent.
Let's remove that and have only one request for all analyses. I believe the current APIs will support the behavior.
Could you check that?
If not, let's create a separate issue for it.

subinasr (approved):

PR #2114, opened 2021-09-03
Create tagging widget for Organigram

### LINKED ISSUE #1815: Create tagging widget for Organigram
## Acceptance Criteria
- The organigram should show a list of selected organigram items on a multi-select list

- [Figma Link](https://www.figma.com/proto/v9SGOBUqs5g5FSCjjwzIqR/DEEP---Wireframes?page-id=0%3A1&node-id=5020%3A8682&viewport=316%2C48%2C0.38&scaling=scale-down-width&starting-point-node-id=4051%3A442&show-proto-sidebar=1&hide-ui=1)

samshara:
- fixes  #1815

## Changes

* Create tagging widget for Organigram

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [ ] permission checks
- [ ] translations

### DIFF
--- a/app/components/OrganigramInput/index.tsx
+++ b/app/components/OrganigramInput/index.tsx
@@ -63,9 +63,9 @@ function OrganigramInput<N extends string>(props: Props<N>) {
         </Card>
     ), [isSelected]);
 
-    const data = useMemo(() => transformData(options as OrganigramDatum), [options]);
+    const data = useMemo(() => options && transformData(options as OrganigramDatum), [options]);
 
-    if (!options) {
+    if (!data) {
         return null;
     }
 
--- a/app/components/OrganigramInput/index.tsx
+++ b/app/components/OrganigramInput/index.tsx
@@ -1,0 +1,69 @@
+import React, { useCallback } from 'react';
+import { _cs } from '@togglecorp/fujs';
+import { PartialForm } from '@togglecorp/toggle-form';
+import OrgTree from 'react-org-tree';
+import {
+    Card,
+} from '@the-deep/deep-ui';
+import styles from './styles.css';
+
+import { OrganigramValue, OrganigramDatum } from '#types/newAnalyticalFramework';
+
+interface Props<N extends string> {
+    name: N;
+    value: OrganigramValue | null | undefined;
+    onChange: (value: OrganigramValue | undefined, name: N) => void,
+    options: PartialForm<OrganigramDatum> | null | undefined;
+    readOnly?: boolean;
+    disabled?: boolean;
+}
+
+function OrganigramInput<N extends string>(props: Props<N>) {
+    const {
+        name,
+        value,
+        onChange,
+        disabled,
+        readOnly,
+        options,
+    } = props;
+
+    const handleClick = useCallback((_: Event, data: OrganigramDatum) => {
+        if (value?.some((v) => v === data.clientId)) {
+            onChange(value?.filter((v) => v !== data.clientId), name);
+        } else {
+            onChange([...(value ?? []), data.clientId], name);
+        }
+    }, [value, onChange, name]);
+
+    const isSelected = useCallback((d: OrganigramDatum) => (
+        value?.find((v) => v === d.clientId)
+    ), [value]);
+
+    const renderContent = useCallback((data: OrganigramDatum) => (
+        <Card
+            className={_cs(styles.card, isSelected(data) && styles.selected)}
+        >
+            { data.label }
+        </Card>
+    ), [isSelected]);
+
+    if (!options) {
+        return null;
+    }
+
+    return (
+        <OrgTree
+            data={options}
+            vertical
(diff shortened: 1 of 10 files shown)


app/components/OrganigramInput/styles.css
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    .card {
+        display: flex;
+        flex-grow: 1;
+        padding: var(--dui-spacing-medium);
+    }
+
+    .selected {
+        border: 1px solid var(--dui-color-accent);
```
AdityaKhatri:
```suggestion
        border: var(--dui-width-separator-thin) solid var(--dui-color-accent);
```

AdityaKhatri (changes requested):

AdityaKhatri (commented):
The imported library doesn't have a type definition or I couldn't find one.
Should we write some kind of type declaration for it or should we look for an alternative.
What do you think @samshara @tnagorra ?

app/components/OrganigramInput/styles.css
```diff
@@ -0,0 +1,31 @@
...
+    .disabled {
+        :global {
+            .org-tree-node-label {
+                pointer-events: none;
+            }
+        }
+    }
+}
+
+
```
tnagorra:
```suggestion
```

app/components/OrganigramInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    ), [value]);
+
+    if (!options) {
+        return null;
+    }
+
+    return (
+        <div
+            className={_cs(styles.container, className)}
+        >
```
tnagorra: Do we need a container div?

app/components/OrganigramInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <OrgTree
+                data={options}
+                vertical
+                renderContent={(data: OrganigramDatum) => (
+                    <Card
+                        className={_cs(styles.card, isSelected(data) && styles.selected)}
+                    >
+                        { data.label }
+                    </Card>
+                )}
```
tnagorra: We should use useCallback for renderContent value

app/components/OrganigramInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        if (value?.some((v) => v === data.clientId)) {
+            onChange(value?.filter((v) => v !== data.clientId), name);
+        } else {
+            onChange(([] as string[]).concat(value ?? [], data.clientId), name);
+        }
+    }, [value, onChange, name]);
+
+    const isSelected = useCallback((d: OrganigramDatum) => (
+        value?.some((v) => v === d.clientId)
+    ), [value]);
```
tnagorra: Shouldn't this just be a `value?.find` ?

app/components/OrganigramInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        disabled,
+        readOnly,
+        options,
+    } = props;
+
+    const handleClick = useCallback((_: Event, data: OrganigramDatum) => {
+        if (value?.some((v) => v === data.clientId)) {
+            onChange(value?.filter((v) => v !== data.clientId), name);
+        } else {
+            onChange(([] as string[]).concat(value ?? [], data.clientId), name);
```
tnagorra: let's prefer this syntax `[...(value ?? []), data.clientId]`

app/components/OrganigramInput/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        name,
+        value,
+        onChange,
+        disabled,
+        readOnly,
+        options,
+    } = props;
+
+    const handleClick = useCallback((_: Event, data: OrganigramDatum) => {
+        if (value?.some((v) => v === data.clientId)) {
```
tnagorra, line 47: Can there be multiple data.clientId in `value`? Why are we using some instead of find?

tnagorra (changes requested):

app/components/OrganigramInput/index.tsx
samshara: boolean value needs to be returned so array.some is more semantic here.

app/components/OrganigramInput/index.tsx
samshara: yes there will be multiple values in `value` props. The value props contains list of `clientId` right now. Maybe later we need to pass idSelector as there will be `id` passed from the server. Isn't `array.some` more semantic to check if item exists in an array and use it inside `if` condition.

AdityaKhatri (changes requested):
Need to fix typing issues and write type declarations for the newly added library

AdityaKhatri (approved):

PR #2119, opened 2021-09-07
Feature/always show non empty filter

### LINKED ISSUE #2080: When filters are collapsed, the inputs  should not be hidden if they have values
This applies to all forms which can be collapsed.

samshara:
- fixes #2080

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [x] permission checks
- [x] translations

### DIFF
--- a/app/utils/common.ts
+++ b/app/utils/common.ts
@@ -1,6 +1,9 @@
 import { memo, ReactElement } from 'react';
 
 import {
+    isNaN,
+    isObject,
+    isList,
     isDefined,
     isNotDefined,
     compareNumber,
@@ -119,4 +122,28 @@ export function joinList(
             ];
         },
     ).flat();
+}
+
+export function hasNoData(obj: unknown): boolean {
+    if (obj === undefined || obj === null || isNaN(obj)) {
+        return true;
+    }
+
+    if (isList(obj)) {
+        if (obj.length <= 0) {
+            return true;
+        }
+        return obj.every((e) => hasNoData(e));
+    }
+
+    if (isObject(obj)) {
+        if (Object.keys(obj).length <= 0) {
+            return true;
+        }
+        return Object.values(obj).every(
+            (value) => hasNoData(value),
+        );
+    }
+
+    return false;
 }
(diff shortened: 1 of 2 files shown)


app/views/Project/Tagging/Sources/SourcesFilter/index.tsx
```diff
@@ -219,7 +223,10 @@ function SourcesFilter(props: Props) {
                 <MultiSelectInput
                     className={_cs(
                         styles.input,
-                        !showContent && styles.hidden,
+                        (!(
+                            Array.isArray(value.authoringOrganizationTypes)
+                            && value.authoringOrganizationTypes.length > 0
+                        ) && !showContent) && styles.hidden,
```
tnagorra: We should define a function that should work for objects, arrays and literals.

tnagorra (changes requested):

app/utils/common.ts
```diff
@@ -120,3 +123,25 @@ export function joinList(
...
+export function hasNoData(obj: unknown): boolean {
+    if (obj === undefined || obj === null || isNaN(obj)) {
+        return true;
+    }
+
+    if (isList(obj)) {
+        if (obj.length <= 0) {
+            return true;
+        }
+        return false;
```
tnagorra:
```suggestion
        return obj.length <= 0;
```

app/utils/common.ts
```diff
@@ -120,3 +123,25 @@ export function joinList(
...
+    }
+
+    if (isObject(obj)) {
+        if (Object.keys(obj).length <= 0) {
+            return true;
+        }
+        return false;
```
tnagorra:
```suggestion
        return Object.keys(obj).length <= 0;
```

app/utils/common.ts
```diff
@@ -120,3 +123,25 @@ export function joinList(
...
+    }
+
+    return false;
```
tnagorra, line 148: We should also handle string.
Most of the time "" is also treated as empty filter.

We should also think about cases like these where it get's a bit fuzzy:
```
{
    startDate: '',
    endDate: '',
}
```

tnagorra (changes requested):

tnagorra (dismissed):

PR #2164, opened 2021-09-29
Implement clone functionality for widgets

### LINKED ISSUE #1810: Create an API to clone widget
## Acceptance Criteria
* Common interface should be introduced for cloning

### LINKED ISSUE #2167: Implement widget clone for all widgets

samshara:
- fixes #1810
- fixes #2167

## Changes

* Add clone function for each widget type

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [x] permission checks
- [x] translations

### DIFF
--- a/app/views/AnalyticalFramework/utils.ts
+++ b/app/views/AnalyticalFramework/utils.ts
@@ -51,7 +51,17 @@ function cloneScaleWidget(widget: ScaleWidget) {
     });
 }
 
-function cloneSelectionWidget(widget: MultiSelectWidget | SingleSelectWidget) {
+function cloneMultiSelectionWidget(widget: MultiSelectWidget) {
+    return cloneWidgetSuperficially({
+        ...widget,
+        properties: {
+            ...widget.properties,
+            options: widget.properties.options.map(cloneOption),
+        },
+    });
+}
+
+function cloneSingleSelectionWidget(widget: SingleSelectWidget) {
     return cloneWidgetSuperficially({
         ...widget,
         properties: {
@@ -116,7 +126,7 @@ function cloneMatrix2dWidget(widget: Matrix2dWidget) {
 // eslint-disable-next-line import/prefer-default-export
 export function cloneWidget(
     widget: Widget,
-) {
+): Widget | undefined {
     switch (widget.widgetId) {
         case 'DATE':
             return cloneWidgetSuperficially(widget);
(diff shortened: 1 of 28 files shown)


app/views/AnalyticalFramework/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+    Matrix2dWidget,
+} from '#types/newAnalyticalFramework';
+import { PartialWidget } from '#components/framework/AttributeInput';
+
+function clone(widget: PartialWidget) {
+    return ({
+        ...widget,
+        title: `${widget.title}-cloned`,
+        key: randomString(),
+        clientId: randomString(),
```
tnagorra: These two may need to have the same value.
@thenav56 what do you think?

app/views/AnalyticalFramework/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+        id: undefined,
```
tnagorra: We should be able to re-use this logic for other widgets a well.
Something like the example below:

```javascript
// NOTE: we need to extend only the basic properties that we are modifying

function cloneWidgetSuperficially<T extends BaseWidget>(value: T): T {
    const randKey = randomString();
    return {
        ...value,
        key: randKey,
        clientId: randKey,
        id: undefined,
        title: `${value.title}: Cloned`
    };
}
```

app/views/AnalyticalFramework/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+        case 'ORGANIGRAM':
+            return cloneOrganigramWidget(widget);
+        case 'MATRIX1D':
+            return cloneMatrix1dWidget(widget);
+        case 'MATRIX2D':
+            return cloneMatrix2dWidget(widget);
+        case 'TEXT':
+            return clone(widget);
+        default:
+            return undefined;
```
tnagorra, line 158: Haven't we used this utiltiy on primary tagging and secondary tagging page?

tnagorra (changes requested):

app/views/AnalyticalFramework/utils.ts
thenav56: Are we now using key instead of clientId? Or both?

Not sure where this is used. But if we have both, then we can easily remove clientId in the future (Currently server uses clientId for widget properties data)

tnagorra: I have started work on PR to convert all "persistent" keys to be named "key".

One problem is that we have "id", "clientId" and "key" on widget level.
We could do away with "key", but don't know how its used on the server

tnagorra (approved):

AdityaKhatri (approved):

PR #2354, opened 2022-01-25
Replace previous logic for switching section

priyesh777:
## Changes:
- Fix issue regarding the section tab not changing to latest (added or edited) in framework details 

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] codegen errors
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [ ] permission checks
- [ ] translations

### DIFF
--- a/app/views/AnalyticalFramework/PrimaryTagging/index.tsx
+++ b/app/views/AnalyticalFramework/PrimaryTagging/index.tsx
@@ -130,8 +130,11 @@ function PrimaryTaggingInput<K extends string>(props: PrimaryTaggingInput<K>) {
         (value: Section[]) => {
             setTempSections(undefined);
             setSections(value, name);
-            if (sectionToEdit) {
+            if (value?.length > 0 && value.find((tab) => tab.clientId === sectionToEdit)) {
                 setSelectedSection(sectionToEdit);
+            } else {
+                const firstTab = value?.length > 0 ? value[0].clientId : undefined;
+                setSelectedSection(firstTab);
             }
         },
         [setSections, name, sectionToEdit],
--- a/app/views/AnalyticalFramework/PrimaryTagging/index.tsx
+++ b/app/views/AnalyticalFramework/PrimaryTagging/index.tsx
@@ -119,6 +119,7 @@ function PrimaryTaggingInput<K extends string>(props: PrimaryTaggingInput<K>) {
     const handleSectionsEditCancel = useCallback(
         () => {
             setTempSections(undefined);
+            setSectionToEdit(undefined);
         },
         [],
     );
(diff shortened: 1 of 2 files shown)


app/views/AnalyticalFramework/PrimaryTagging/index.tsx
```diff
@@ -432,7 +436,7 @@ function PrimaryTaggingInput<K extends string>(props: PrimaryTaggingInput<K>) {
                                         analyzeErrors(error?.[section.clientId]) && styles.errored,
                                     )}
                                     title={section.tooltip}
-                                    // FIXME: use strings
+                                // FIXME: use strings
```
tnagorra: Please do not change whitespace

app/views/AnalyticalFramework/PrimaryTagging/index.tsx
```diff
@@ -127,8 +128,11 @@ function PrimaryTaggingInput<K extends string>(props: PrimaryTaggingInput<K>) {
         (value: Section[]) => {
             setTempSections(undefined);
             setSections(value, name);
+            if (sectionToEdit) {
+                setSelectedSection(sectionToEdit);
+            }
```
tnagorra: We should not use sectionToEdit to select selectionSection.

On each section change, we should see if selectedSection is removed.
If it's removed, we should set other section as selected (can be the first or the nearest element)

We should also do this check on section cancel, which should restore the previous section if currently selected section is no longer there.

app/views/AnalyticalFramework/PrimaryTagging/index.tsx
```diff
@@ -127,8 +128,14 @@ function PrimaryTaggingInput<K extends string>(props: PrimaryTaggingInput<K>) {
         (value: Section[]) => {
             setTempSections(undefined);
             setSections(value, name);
+            if (value?.length > 0 && value.find((tab) => tab.clientId === sectionToEdit)) {
+                setSelectedSection(sectionToEdit);
+            } else {
+                const firstTab = value?.length > 0 ? value[0].clientId : undefined;
+                setSelectedSection(firstTab);
+            }
```
tnagorra: Are we sure value has at least one element?

tnagorra (changes requested):

app/views/AnalyticalFramework/PrimaryTagging/index.tsx
priyesh777: Yes, the sections element will have at least one element by default.

tnagorra: Let's change this logic so that we check
1. value.length is less than zero then set undefined as selected section
2. value's clientId is equal to sectionToEdit then set that as selected section
3. set first element as selected section

tnagorra (changes requested):

PR #2446, opened 2022-04-26
Fix/add latest fetched

AdityaKhatri:
## Changes

* Add latest fetched at dates


app/views/Project/Tagging/Sources/UnifiedConnectorModal/ProjectConnectorsPane/index.tsx
```diff
@@ -35,11 +39,22 @@ function ConnectorItem(props: ConnectorItemProps) {
...
 
+    const latestFetchedAt = useMemo(() => {
+        if (!sources) {
+            return undefined;
+        }
+        const lastFetchedDates = sources
+            .map((source) => source.lastFetchedAt)
+            .filter(isDefined)
+            .sort(compareDate);
+        return lastFetchedDates[0];
```
tnagorra: Let's add a note that we are just getting a large/small item.
This is a very in-efficient way to find the largest/smallest item.

tnagorra (changes requested):

PR #2549, opened 2022-07-21
Add user delete functionality

AdityaKhatri:
depends on https://github.com/the-deep/server/pull/1118

## Changes

* Add the ability to delete user

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] codegen errors
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [x] permission checks
- [x] translations

### DIFF
--- a/app/components/selections/ProjectUserSelectInput/index.tsx
+++ b/app/components/selections/ProjectUserSelectInput/index.tsx
@@ -44,7 +44,12 @@ const PROJECT_USERS = gql`
     }
 `;
 
-export type BasicProjectUser = NonNullable<NonNullable<NonNullable<NonNullable<ProjectUserQuery['project']>['userMembers']>['results']>[number]>['member'];
+export type BasicProjectUser = {
+    id: string;
+    emailDisplay: string;
+    displayName?: string | null | undefined;
+};
+
 const keySelector = (d: BasicProjectUser) => d.id;
 const labelSelector = (d: BasicProjectUser) => d.displayName ?? ' ';
 
--- a/app/views/MyProfile/styles.css
+++ b/app/views/MyProfile/styles.css
@@ -89,6 +89,10 @@
                         gap: var(--dui-spacing-medium);
                     }
                 }
+
+                .delete-button {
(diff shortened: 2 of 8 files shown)


app/views/MyProfile/styles.css
```diff
@@ -89,6 +89,10 @@
                         gap: var(--dui-spacing-medium);
                     }
                 }
+
+                .delete-button {
+                    color: var(--dui-color-danger);
+                }
```
subinasr, line 95: I see both 'cancel' and the 'delete user' button is assigned the same classname. Are you sure both buttons should appear the same? Or is this a mistake?

app/components/selections/ProjectUserSelectInput/index.tsx
```diff
@@ -44,7 +44,12 @@ const PROJECT_USERS = gql`
     }
 `;
 
-export type BasicProjectUser = NonNullable<NonNullable<NonNullable<NonNullable<ProjectUserQuery['project']>['userMembers']>['results']>[number]>['member'];
+export type BasicProjectUser = {
+    id: string;
+    emailDisplay: string;
+    displayName?: string | null | undefined;
+};
+
```
subinasr, line 52: Can you tell me why you've assigned a type like this instead of using the type from graphql itself?

subinasr (changes requested):

app/views/MyProfile/styles.css
AdityaKhatri: This has been fixed.

app/components/selections/ProjectUserSelectInput/index.tsx
AdityaKhatri: Yeah, while we can infer from the type, we were using this type outside as well and we were transforming items. The types exactly didn't match.

subinasr: I don't understand. We can still export the types and shouldn't it still match?

subinasr (approved):

Issue #2727, opened 2023-08-09
Use graphql to fetch list of analysis frameworks in projectEdit page


subinasr:
### Overview
Currently, the list of analysis frameworks shown on the projectEdit page is fetched through REST API. Use graphql to fetch the list.

### Acceptance Criteria
- The list should be paginated and should handle the filter to show recently used frameworks.
- This change may trigger type issues, which should be properly handled.

### Screenshots

![Image](https://github.com/the-deep/client/assets/51225380/bd8ee9b9-90b8-4362-a7cf-6f40d8e8487e)

PR #2765, opened 2023-09-22
Use graphql to fetch analysis Frameworks List

RobertKoikai:
- Addresses Use graphql to fetch list of analysis frameworks in projectEdit page #2727 
- Depends on :  (https://github.com/the-deep/server/pull/1369 )

## Changes

* Update the projectEdit  Framework page to fetch frameworks using graphql, 
* Remove all  REST API dependencies 
* Fix type related issues

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] codegen errors
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [ ] permission checks
- [ ] translations

### DIFF
--- a/app/views/ProjectEdit/Framework/index.tsx
+++ b/app/views/ProjectEdit/Framework/index.tsx
@@ -46,20 +46,24 @@ interface FrameworkMini {
     title: string;
     createdAt: string;
 }
-const frameworkKeySelector = (d: FrameworkMini) => parseInt(d.id, 10);
+
+const frameworkKeySelector = (d: FrameworkMini) => d.id;
 
 export const PROJECTS_AFS = gql`
     query ProjectAnalysisFrameworks(
         $isCurrentUserMember: Boolean
         $search: String,
     ) {
-        analysisFrameworks(search: $search , isCurrentUserMember: $isCurrentUserMember) {
+        analysisFrameworks(
+            search: $search , 
+            isCurrentUserMember: $isCurrentUserMember
+            ) {
          results {
-                    title
-                    id
-                    createdAt
-                }
-                totalCount
+            title
+            id
+            createdAt
+         }
+         totalCount
         }
     }
 `;
@@ -172,28 +176,14 @@ function ProjectFramework(props: Props) {
     } = useForm(schema, defaultFormValue);
 
     const delayedValue = useDebouncedValue(value);
-    useEffect(() => {
-        setOffset(0);
-        setFrameworkList([]);
-    }, [delayedValue]);
-    const analysisFrameworkVariables = useMemo(() => {
-        let isCurrentUserMember;
-        if (delayedValue.relatedToMe === 'true') {
-            isCurrentUserMember = true;
-        } else if (delayedValue.relatedToMe === 'false') {
-            isCurrentUserMember = false;
-        } else {
-            isCurrentUserMember = undefined;
-        }
-        return {
-            isCurrentUserMember,
-            search: delayedValue.search,
-            offset,
-        };
-    }, [delayedValue]);
+    const analysisFrameworkVariables = useMemo(() => ({
+        isCurrentUserMember: delayedValue.relatedToMe === 'true' ? true : undefined,
+        search: delayedValue.search,
+        offset,
+    }), [delayedValue, offset]);
     const {
         previousData,
-        data: projectUsersResponse = previousData,
+        data: projectAflist = previousData,
         loading: frameworksGetPending,
         refetch: retriggerGetFrameworkListRequest,
     } = useQuery<ProjectAnalysisFrameworksQuery, ProjectAnalysisFrameworksQueryVariables>(
@@ -203,7 +193,7 @@ function ProjectFramework(props: Props) {
         },
     );
 
-    const frameworksRendererParams = useCallback((key: number, data: FrameworkMini) => ({
+    const frameworksRendererParams = useCallback((key: string, data: FrameworkMini) => ({
         itemKey: String(key),
         title: data.title,
         createdAt: data.createdAt,
@@ -229,11 +219,9 @@ function ProjectFramework(props: Props) {
         offset,
     ]);
 
-    const totalFrameworksCount = projectUsersResponse?.analysisFrameworks?.totalCount ?? 0;
-    const AfList: FrameworkMini[] = projectUsersResponse?.analysisFrameworks?.results || [];
+    const totalFrameworksCount = projectAflist?.analysisFrameworks?.totalCount ?? 0;
+    const afList: FrameworkMini[] = projectAflist?.analysisFrameworks?.results || [];
 
-    console.log('type one', typeof frameworkList);
-    console.log('type two >> ', projectUsersResponse?.analysisFrameworks?.results);
     return (
         <div className={_cs(styles.framework, className)}>
             <div className={styles.leftContainer}>
@@ -261,7 +249,7 @@ function ProjectFramework(props: Props) {
                         className={styles.frameworkList}
                         pending={frameworksGetPending}
                         errored={false}
-                        data={AfList}
+                        data={afList}
                         keySelector={frameworkKeySelector}
                         renderer={Item}
                         filtered={isFiltered(value)}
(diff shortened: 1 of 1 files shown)


app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -214,8 +229,11 @@ function ProjectFramework(props: Props) {
         offset,
     ]);
 
-    const totalFrameworksCount = frameworksResponse?.count ?? 0;
+    const totalFrameworksCount = projectUsersResponse?.analysisFrameworks?.totalCount ?? 0;
+    const AfList: FrameworkMini[] = projectUsersResponse?.analysisFrameworks?.results || [];
 
+    console.log('type one', typeof frameworkList);
+    console.log('type two >> ', projectUsersResponse?.analysisFrameworks?.results);
```
subinasr: Let's remove this.

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -22,30 +22,47 @@ import {
...
+        $isCurrentUserMember: Boolean
+        $search: String,
+    ) {
+        analysisFrameworks(search: $search , isCurrentUserMember: $isCurrentUserMember) {
+         results {
+                    title
+                    id
+                    createdAt
+                }
+                totalCount
```
AdityaKhatri: fix indentation

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -22,30 +22,47 @@ import {
...
 interface FrameworkMini {
-    id: number;
+    id: string;
     title: string;
     createdAt: string;
 }
-
-const frameworkKeySelector = (d: FrameworkMini) => d.id;
-
+const frameworkKeySelector = (d: FrameworkMini) => parseInt(d.id, 10);
```
AdityaKhatri: Why parseInt?

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -156,37 +172,36 @@ function ProjectFramework(props: Props) {
...
+        let isCurrentUserMember;
+        if (delayedValue.relatedToMe === 'true') {
+            isCurrentUserMember = true;
+        } else if (delayedValue.relatedToMe === 'false') {
+            isCurrentUserMember = false;
+        } else {
+            isCurrentUserMember = undefined;
+        }
+        return {
+            isCurrentUserMember,
```
AdityaKhatri: We should be able to directly pass relatedToMe as isCurrentUserMember. Can you check?

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -156,37 +172,36 @@ function ProjectFramework(props: Props) {
...
+            search: delayedValue.search,
+            offset,
```
AdityaKhatri: The order is missing.

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -156,37 +172,36 @@ function ProjectFramework(props: Props) {
     } = useForm(schema, defaultFormValue);
 
     const delayedValue = useDebouncedValue(value);
-
     useEffect(() => {
         setOffset(0);
         setFrameworkList([]);
     }, [delayedValue]);
```
AdityaKhatri: While you are here, this should also be removed. Can you make sure to update frameworkList and offset as soon as `value` is changed?

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -214,8 +229,11 @@ function ProjectFramework(props: Props) {
         offset,
     ]);
 
-    const totalFrameworksCount = frameworksResponse?.count ?? 0;
+    const totalFrameworksCount = projectUsersResponse?.analysisFrameworks?.totalCount ?? 0;
+    const AfList: FrameworkMini[] = projectUsersResponse?.analysisFrameworks?.results || [];
```
AdityaKhatri:
```suggestion
    const frameworkList: FrameworkMini[] = projectUsersResponse?.analysisFrameworks?.results || [];
```
Ts variables should always be camelcase.

app/views/ProjectEdit/Framework/index.tsx
AdityaKhatri: Please make sure there are no console logs left before you sent the PR.
My recommendation is doing a self code review as soon as the PR is sent.

AdityaKhatri (changes requested):

app/views/ProjectEdit/Framework/index.tsx
subinasr: For future, we have a checklist while creating PRs, checking those boxes only after confirming it is a recommended practice.

RobertKoikai:
@AdityaKhatri ,

The PR has been updated.

Regarding the order please note that the backend for this is also modified to show frameworks that have been used to tag an entry within six months and list the filtered options with the most recently used framework on top.(See PR : https://github.com/the-deep/server/pull/1369).

cc ; @subinasr

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -43,9 +46,26 @@ interface FrameworkMini {
...
 
+export const PROJECTS_AFS = gql`
+    query ProjectAnalysisFrameworks(
+        $isCurrentUserMember: Boolean
+        $search: String,
+    ) {
+        analysisFrameworks(
+            search: $search , 
+            isCurrentUserMember: $isCurrentUserMember
+            ) {
```
AdityaKhatri: Fix indentation

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -156,37 +175,22 @@ function ProjectFramework(props: Props) {
...
-        setFrameworkList([]);
-    }, [delayedValue]);
-
-    const queryForFrameworks = useMemo(() => ({
-        fields: ['id', 'title', 'created_at'],
-        offset,
-        order: 'created_at',
-        limit: 10,
+    const analysisFrameworkVariables = useMemo(() => ({
+        isCurrentUserMember: delayedValue.relatedToMe === 'true' ? true : undefined,
```
AdityaKhatri:
```suggestion
        isCurrentUserMember: delayedValue.relatedToMe,
```
Does this not work?

app/views/ProjectEdit/Framework/index.tsx
```diff
@@ -214,8 +218,12 @@ function ProjectFramework(props: Props) {
...
     ]);
 
-    const totalFrameworksCount = frameworksResponse?.count ?? 0;
-
+    const totalFrameworksCount = projectAflist?.analysisFrameworks?.totalCount ?? 0;
+    const afList: FrameworkMini[] = (projectAflist?.analysisFrameworks?.results || [])
+        .map((item) => ({
+            ...item,
+            id: Number(item.id),
+        }));
```
AdityaKhatri: Lets not cast id to number.
After adopting graphql, all the id fields are string as well.
As this was in REST before, we used number to save selectedFramework, but it was eventually cast to string while checking whether its active or not.
Lets fix that there, and not cast.

Basically I recommend not casting it and fixing other places it causes issue to.

AdityaKhatri (changes requested):

app/views/ProjectEdit/Framework/index.tsx
RobertKoikai: @AdityaKhatri ,
By  "fixing other places it causes issue to" does this mean fix the type issues in ListView in **deep-ui** ?
.

app/views/ProjectEdit/Framework/index.tsx
RobertKoikai: @AdityaKhatri ,

To return ALL the analytical frameworks, relatedToMe should be neither true nor false. 

True returns **only** the AFs belonging to me .
False returns **only** the AFs that do not belong to me.

app/views/ProjectEdit/Framework/index.tsx
AdityaKhatri: No, type issues will be in the rendererParams of ListView, so we need to fix it there. Let me know if a call is necessary.

app/views/ProjectEdit/Framework/index.tsx
AdityaKhatri: Yes, and by default  relatedToMe is false or undefined? If its undefined then there shouldn't be any issue right?

AdityaKhatri (approved):

Issue #2774, opened 2023-10-12
Allow font-size changes in Simplified texts.


subinasr:
## Issue
Simplified texts are difficult to read due to their small font-size.

## Acceptance Criteria
Add a '+/-' button to increase or decrease the text size.

PR #2806, opened 2023-12-26
Add text zoom in simplified text

subinasr:
- Addresses #2774 

## Changes

* Add text zoom feature in simplified text view

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] codegen errors
- [x] typos
- [x] unwanted comments
- [x] conflict markers

### DIFF
--- a/app/views/EntryEdit/LeftPane/SimplifiedTextView/styles.css
+++ b/app/views/EntryEdit/LeftPane/SimplifiedTextView/styles.css
@@ -14,20 +14,29 @@
 
     .simplified-text {
         line-height: 1.5;
-        &.extra-small{
-            font-size: var(--dui-font-size-extra-small);
+        --base-font-size: var(--dui-font-size-medium);
+        --diff: 2px;
+
+        &.x-small{
+            font-size: var(--base-font-size);
         }
         &.small{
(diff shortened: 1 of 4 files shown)


app/views/EntryEdit/LeftPane/SimplifiedTextView/styles.css
```diff
@@ -12,6 +12,28 @@
...
     }
 
+    .simplified-text {
+        line-height: 1.5;
+        --base-font-size: var(--dui-font-size-medium);
+        --diff: 2px;
+
+        &.extra-small{
+            font-size: calc(var(--base-font-size) - 2 * var(--diff));
+        }
```
AdityaKhatri: I think the font we have should be the smallest allowed font and only allow increments. What do you think?

AdityaKhatri (changes requested):

app/views/EntryEdit/LeftPane/SimplifiedTextView/styles.css
subinasr: That is what the process is now. Take a look please.

AdityaKhatri (approved):

PR #2976, opened 2024-07-30
Fix attachment issue

barshathakuri:
## Changes

* Fix attachment issue in primary and secondary tab.
* Move attachment state

- [x] build works
- [x] eslint issues
- [x] typescript issues
- [x] codegen errors
- [x] typos
- [x] unwanted comments
- [x] conflict markers

## This PR contains valid:

- [x] permission checks
- [x] translations

### DIFF
--- a/app/components/LeftPaneEntries/index.tsx
+++ b/app/components/LeftPaneEntries/index.tsx
@@ -149,6 +149,7 @@ interface Props {
     entries: EntryInput[] | undefined | null;
     activeEntry?: string;
     onEntryClick?: (entryId: string) => void;
+    onAttachmentClick: (attachment: LeadPreviewAttachmentType) => void;
     onExcerptChange?: (entryClientId: string, newExcerpt: string | undefined) => void;
     onEntryDelete?: (entryClientId: string) => void;
     onEntryRestore?: (entryClientId: string) => void;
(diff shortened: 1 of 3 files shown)


shreeyash07 (approved):

app/components/LeftPaneEntries/index.tsx
```diff
@@ -396,23 +398,6 @@ function LeftPane(props: Props) {
...
-                excerpt: '',
-                droppedExcerpt: '',
-            });
-            setLeadAttachmentsMap((oldValue) => ({
-                ...oldValue,
-                [attachment.id]: attachment,
-            }));
-        }
-    }, [leadId, onEntryCreate]);
-
```
subinasr, line 415: Is there a reason we're handling this in 'EntryEdit' and not here?

subinasr (changes requested):
Please fix all type issues.

app/components/LeftPaneEntries/index.tsx
barshathakuri: @subinasr  By adding handleAttachmentClick to the LeftPane Entries, the leadId is utilized in both the secondary and review tabs. Moving it to entryEdit makes the leadId accessible across all tabs, which explains why the image was only visible in the primary tab previously.

